PR #1, opened 2019-09-25
TRON-39 - tron skeleton

twtaylorbitgo:
This is the tron skeleton, eventually we want to remove all the net specific code.


test/unit/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    signature:
+     [ ] 
+    };
+
+  it('be able to convert hex to bytes', () => {
+    const ba = Tron.hexStr2byteArray(hex);
+    should.deepEqual(ba, arr);
+  });
+
+  it('be able to convert hex to bytes', () => {
```
glowang: The description of this test does not match its behaviors, and it also looks like it is trying to do the same as the above test

src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+      console.log('There was an error decoding the transfer contract in the transaction.');
+      throw e;
+    }
+
+    assert(transferContract.ownerAddress);
+    assert(transferContract.toAddress);
+    assert(transferContract.hasOwnProperty('amount'));
+
+    // deserialize attributes
+    const ownerAddress = Tron.getBase58AddressFromByteArray(Tron.hexStr2byteArray(Buffer.from(transferContract.ownerAddress, 'base64').toString('hex')));
```
glowang: Is it supposed to be base58...?

src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    const toAddress = Tron.getBase58AddressFromByteArray(Tron.hexStr2byteArray(Buffer.from(transferContract.toAddress, 'base64').toString('hex')));
```
glowang: same question as above

test/unit/index.ts
twtaylorbitgo: ah good point. i'll make the name more meaningful

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      amount,
+    };
+  }
+
+  static decodeAccountPermissionUpdateContract(base64: string): ParsedAccountUpdatePermission {
+    throw new Error('Not implemented yet.');
+  }
+
+  static generateAccount(): Account {
+    return tronweb.utils.accounts.generateAccount();
```
tylerlevine: How come you're wrapping all these one liner functions from tron lib up without any other logic?

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+export default class Tron {
+
+  constructor(coinName: string) {
+    // TODO: this will probably need to become more useful if we need to decode if based on network
+  }
+
+  static decodeContract(base64EncodedHex: string): DecodedContract | ParsedAccountUpdatePermission {
+    const rawContract = this.getRawTransferContract(base64EncodedHex);
+    // there should not be multiple contracts in this data
+    assert(rawContract.contractList.length === 1);
```
tylerlevine: Please throw an Error with message, otherwise someone will have to debug this: `AssertionError [ERR_ASSERTION]: false == true`

In general, `assert()` should be avoided in non-test code, and even then the methods from shouldjs should be used.

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  static getRawTransferContract(base64EncodedHex: string): transferContractRaw {
+    const bytes = Buffer.from(base64EncodedHex, 'hex');
+    let raw;
+    // we need to decode our raw_data_hex field first
+    try {
+      raw = tronproto.Transaction.raw.deserializeBinary(bytes).toObject();
+    } catch (e) {
+      console.log('There was an error decoding the initial raw_data_hex from the serialized tx.');
```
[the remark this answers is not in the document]
twtaylorbitgo: ack

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+const tronproto = require('./protobuf/tron_pb');
+const contractproto = require('./protobuf/Contract_pb');
+
+import * as assert from 'assert';
+
+// this is a construct unique to tron
+export type ByteArray = number[];
+export type Transaction = { txID: string, signature?: string[] };
+
+export default class Tron {
```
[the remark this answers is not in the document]
twtaylorbitgo: ack

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+    }
+    return {
+      expiration: raw.expiration,
+      timestamp: raw.timestamp,
+      contractList: raw.contractList,
+    };
+  }
+
+  /** A valid hex string must be a string made of numbers and characters and has an even length.*/
+  static safeTronHexString(str: string): Boolean {
```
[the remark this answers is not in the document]
twtaylorbitgo: what about this?

https://nodejs.org/api/buffer.html#buffer_class_method_buffer_isencoding_encoding

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+   * @param transferHex is the value property of the "parameter" field of contractList[0]
+   * */
+  static decodeTransferContract(transferHex: string): DecodedContract {
+    const contractBytes = Buffer.from(transferHex, 'base64');
+    let transferContract;
+
+    try {
+      transferContract = contractproto.TransferContract.deserializeBinary(contractBytes).toObject();
+    } catch (e) {
+      console.log('There was an error decoding the transfer contract in the transaction.');
```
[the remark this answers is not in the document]
twtaylorbitgo: removed

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+      timestamp: raw.timestamp,
+      contractList: raw.contractList,
+    };
+  }
+
+  /** A valid hex string must be a string made of numbers and characters and has an even length.*/
+  static safeTronHexString(str: string): Boolean {
+    // the second and third condition look unnecessary in Typescript, but since we are eventually running
+    // the javascript version of the library under the hood, they are included.
+    if (!str || typeof(str) !== 'string' || !str.match(/^(0x)?[0-9a-fA-F]*$/)) {
```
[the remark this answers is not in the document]
twtaylorbitgo: ok this should work except I think it's not match, it's test

src/index.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  static hexStr2byteArray(str: string): ByteArray {
+    return tronweb.utils.code.hexStr2byteArray(str);
+  }
+
+  static byteArray2hexStr(arr: ByteArray): string {
+    return tronweb.utils.code.byteArray2hexStr(arr);
+  }
+
+  static getPubKeyFromPriKey(privateKey: ByteArray | Buffer | Uint8Array | string): ByteArray {
```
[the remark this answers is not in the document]
twtaylorbitgo: ach I added a type here

src/index.ts
twtaylorbitgo: I think we don't want to surface the entire tronlib to the user here

src/index.ts
twtaylorbitgo: ah shoot, good point, yeah these asserts should have never made it into live code

twtaylorbitgo:
Note @tylerlevine 's comments have been addressed in the pr #7

PR #2, opened 2019-10-14
TRON-71 Add Helper Function to Account Lib 

glowang:
The explain transaction function needs information about the transaction timestamp and expiration, and we need to add a helper function within index.ts for them.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,16 +8,31 @@ import * as assert from 'assert';
 export type ByteArray = number[];
 export type Transaction = { txID: string, signature?: string[] };
 
-export default class Tron { 
+export default class Tron {
 
   constructor(coinName: string) {
     // TODO: this will probably need to become more useful if we need to decode if based on network
   }
 
-  static parseTransaction(base64EncodedHex: string): ParsedTransaction | ParsedAccountUpdatePermission {
+  static decodeContract(base64EncodedHex: string): DecodedContract | ParsedAccountUpdatePermission {
+    const rawContract = this.getRawTransferContract(base64EncodedHex);
+    // there should not be multiple contracts in this data
+    assert(rawContract.contractList.length === 1);
+
+    // ensure the contract type is supported
+    switch  (rawContract.contractList[0].parameter.typeUrl) {
+      case 'type.googleapis.com/protocol.TransferContract':
+        return this.decodeTransferContract(rawContract.contractList[0].parameter.value);
+      case 'type.googleapis.com/protocol.AccountPermissionUpdateContract':
+        return this.decodeAccountPermissionUpdateContract(rawContract.contractList[0].parameter.value);
+      default:
+        throw new Error('Unsupported contract type');
+    }
+  }
+
+  static getRawTransferContract(base64EncodedHex: string): transferContractRaw {
     const bytes = Buffer.from(base64EncodedHex, 'hex');
     let raw;
-
     // we need to decode our raw_data_hex field first
     try {
       raw = tronproto.Transaction.raw.deserializeBinary(bytes).toObject();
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -40,6 +40,24 @@ export default class Tron {
...
+    try {
+      raw = tronproto.Transaction.raw.deserializeBinary(bytes).toObject();
+    } catch (e) {
+      console.log('There was an error decoding the initial raw_data_hex from the serialized tx.');
+      throw e;
+    }
+    return { 
+      expiration: raw.expiration, 
+      timestamp: raw.timestamp,
+      value: raw.contractList[0].parameter.value,
```
twtaylorbitgo: I would call this transferContractRaw

src/index.ts
```diff
@@ -40,6 +40,24 @@ export default class Tron {
     }
   }
 
+  static getTransactionData(base64EncodedHex: string): TransactionData {
```
Xekyo: `getTransactionData` sounds sort of like a `parseTransaction`-type of deal. Is this not just for some specific metadata? The function is missing documentation of its purpose and parameters.

src/index.ts
```diff
@@ -40,6 +40,24 @@ export default class Tron {
...
+      throw e;
+    }
+    return { 
+      expiration: raw.expiration, 
+      timestamp: raw.timestamp,
+      transferContractRaw: raw.contractList[0].parameter.value,
+    };
+  }
+
+  /** the parameter should not be the hex, but the contract bytes instead */
```
Xekyo: I don't understand the purpose of this comment. If this is a matter of the parameter being named incorrectly, you could do it in this PR in a second commit. If there is a bug that you will not address in this PR, you should file a Jira ticket and add a `// FIXME(TICKET-123) description of what is broken`.

src/index.ts
```diff
@@ -139,3 +157,9 @@ export interface Account {
   publicKey: string;
   privateKey: string;
 }
+
+export interface TransactionData {
```
Xekyo: I would probably name this differently, because it doesn't seem to be the full transaction data but a subset. The class name should be probably more specific.

test/unit/index.ts
```diff
@@ -115,4 +114,10 @@ describe('Tron library should', function() {
     should.equal(parsedTx.amount, amount);
   });
 
+  it ('should return transaction data', () => {
+    const data = Tron.getTransactionData(tx.raw_data_hex);
```
Xekyo: I'm curious, how was the testing data derived? If it used the same code to create the data, it would be a tautological test. I hope the testdata was retrieved from Tron's implementation guidelines/testcases?

Do we have another `parseTransaction` function already that returns this data? If that's the case why are we implementing `getTransactionData` here instead of using the existing `parseTransaction`?

glowang: > I'm curious, how was the testing data derived? If it used the same code to create the data, it would be a tautological test. I hope the testdata was retrieved from Tron's implementation guidelines/testcases?
> 
> Do we have another `parseTransaction` function already that returns this data? If that's the case why are we implementing `getTransactionData` here instead of using the existing `parseTransaction`?

We derived the test data using coin sandbox, which generates data from Tron testnet. We have parseTransaction but it does not return all the information we want, it is a sign that we should name these functions better. Good catch.

src/index.ts
```diff
@@ -8,40 +8,60 @@ import * as assert from 'assert';
 export type ByteArray = number[];
 export type Transaction = { txID: string, signature?: string[] };
 
-export default class Tron { 
+export default class Tron {
```
cooncesean: @argjv I thought that we were going to move coin specific classes/logic to their own modules? ie:

```
- src/
  - trx/
    - trx.ts
  - eth/
     - eth.ts
```

src/index.ts
```diff
@@ -8,40 +8,60 @@ import * as assert from 'assert';
...
 
   constructor(coinName: string) {
     // TODO: this will probably need to become more useful if we need to decode if based on network
   }
 
   static parseTransaction(base64EncodedHex: string): ParsedTransaction | ParsedAccountUpdatePermission {
```
cooncesean: Is this fcn supposed to return anything (like a parsed transaction)? Seems like it would be more appropriately named `validateContract`.

src/index.ts
```diff
@@ -8,40 +8,60 @@ import * as assert from 'assert';
...
-    // ensure the contract type is supported
-    switch  (raw.contractList[0].parameter.typeUrl) {
-      case 'type.googleapis.com/protocol.TransferContract':
-        return this.decodeTransferContract(raw.contractList[0].parameter.value);
-      case 'type.googleapis.com/protocol.AccountPermissionUpdateContract':
-        return this.decodeAccountPermissionUpdateContract(raw.contractList[0].parameter.value);
-      default:
-        throw new Error('Unsupported contract type');    
+  static safeTronHexString(str: string): Boolean {
+    if (!str || typeof(str) !== 'string' || !str.match(/^(0x)?[0-9a-fA-F]*$/)) {
```
cooncesean: Can you add an inline comment describing the conditional and why it's needed?

test/unit/index.ts
```diff
@@ -114,5 +120,4 @@ describe('Tron library should', function() {
     should.equal(parsedTx.ownerAddress, ownerAddress);
     should.equal(parsedTx.amount, amount);
   });
-
-});
\ No newline at end of file
+});
```
cooncesean: Do we need to add any unit tests for the changes made to `parseTransaction` and/or `safeTronHexString`?

glowang: I took the logic for safeTronHexString from Platform 
`exports.safeHexString = function (str, coin) {
  if (!str || typeof(str) !== 'string' || !str.match(/^(0x)?[0-9a-fA-F]*$/)) {
    return undefined;
  }
  if ((str.length % 2) !== 0) {
    return undefined;
  }
  return (coin === 'eth') ? ethUtil.addHexPrefix(str).toLowerCase() : str;
};`

Took out the eth part since we haven't added logic for eth in our library yet

src/index.ts
glowang: what about decodeContract?

Xekyo:
My comments have been addressed, I've dismissed my previous review.

PR #3, opened 2019-10-17
TRON-39 - account lib skeleton

twtaylorbitgo:
This is the skeleton for account lib. It is mostly based off this document.

https://docs.google.com/document/d/17snNmlElYph59maFdbxF_jCnIF3wDa8e3Wf0fJ5fJdU/edit#heading=h.79rp2kie2cj


package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@bitgo/bitgo-account-lib",
+  "version": "1.0.0-rc.0",
```
argjv: This should probably be `0.0.1` and change it to `1.0.0` once we properly add Tron

.gitignore
```diff
@@ -0,0 +1,8 @@
+node_modules/
```
argjv: Add `.idea/` which is added by IntelliJ

package.json
twtaylorbitgo: sure

.gitignore
twtaylorbitgo: k

PR #4, opened 2019-10-18
TRON-39 - create account lib scaffolding

twtaylorbitgo:
This is part of TRON-39. The purpose of this series of commits is to create scaffolding for account-lib's transactionBuilder. The final goal will be base tests, along with Tron as a mock (just stubbed out for now)

### DIFF
--- a/src/coin/baseCoin.ts
+++ b/src/coin/baseCoin.ts
@@ -1,0 +1,30 @@
+import { Network, ITransaction, ISignature, IKey, TransactionType } from "..";
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import BigNumber from "bignumber.js";
+
+export abstract class BaseCoin {
+  public network: Network;
+  protected staticsCoin: Readonly<StaticsBaseCoin>;
+
+  constructor(network: Network) {
+    this.network = network;
+  }
+
+  // Getters
+  abstract get displayName(): string;
+  abstract get maxFrom(): number;
+  abstract get maxDestinations(): number;
+
+  // Validation helpers
+  public abstract validateAddress(address: string): boolean;
+  public abstract validateValue(value: BigNumber): boolean;
+  public abstract validateAddDestination(address: string, value: BigNumber): boolean;
+  public abstract validateAddSender(address: string): boolean;
+
+  // Parse & build
+  public abstract parseTransaction(rawTransaction: any, transactionType: TransactionType): ITransaction;
+  public abstract buildTransaction(transaction: ITransaction, transactionType: TransactionType): ITransaction;
+
+  // Sign
+  public abstract sign(privateKey: IKey, address: string, transaction: ITransaction): ISignature;
+}
--- a/src/coin/baseCoin.ts
+++ b/src/coin/baseCoin.ts
@@ -1,4 +1,4 @@
-import { Network, ITransaction, ISignature, IKey, TransactionType } from "..";
+import { Network, Transaction, Signature, Key, TransactionType } from "..";
 import { coins, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import BigNumber from "bignumber.js";
 
@@ -49,13 +49,13 @@ export abstract class BaseCoin {
    * @param rawTransaction the raw transaction in this case. format determined by the coin
    * @param transactionType the type/direction of the transaction
    */
-  public abstract parseTransaction(rawTransaction: any, transactionType: TransactionType): ITransaction;
+  public abstract parseTransaction(rawTransaction: any, transactionType: TransactionType): Transaction;
 
   /**
    * Build our transaction. Returns the resultant transaction with a completed transaction (or error state)
    * @param transaction 
    */
-  public abstract buildTransaction(transaction: ITransaction): ITransaction;
+  public abstract buildTransaction(transaction: Transaction): Transaction;
 
   /**
    * Sign a transaction. Returns a signature or a signature with invalid state.
@@ -63,5 +63,5 @@ export abstract class BaseCoin {
    * @param address 
    * @param transaction the transaction 
    */
-  public abstract sign(privateKey: IKey, address: string, transaction: ITransaction): ISignature;
+  public abstract sign(privateKey: Key, address: string, transaction: Transaction): Signature;
 }
--- a/src/coin/baseCoin.ts
+++ b/src/coin/baseCoin.ts
@@ -1,30 +1,67 @@
 import { Network, ITransaction, ISignature, IKey, TransactionType } from "..";
-import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import { coins, BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
 import BigNumber from "bignumber.js";
 
 export abstract class BaseCoin {
+  /**
+   * The network this transaction is being built for.
+   */
   public network: Network;
+
+  /**
+   * Our coin from statics
+   */
   protected staticsCoin: Readonly<StaticsBaseCoin>;
 
   constructor(network: Network) {
     this.network = network;
   }
 
-  // Getters
+  /**
+   * The statics fullName of this coin
+   */
   abstract get displayName(): string;
+
+  /**
+   * This is the maximum senders this coin can have for a given transaction.
+   */
   abstract get maxFrom(): number;
+
+  /**
+   * This is the maximum destinations this coin can have for a given transaction.
+   */
   abstract get maxDestinations(): number;
 
-  // Validation helpers
-  public abstract validateAddress(address: string): boolean;
+  /**
+   * Validate an address. Returns false if the address is invalid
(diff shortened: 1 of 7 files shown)


src/coin/baseCoin.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  // Getters
+  abstract get displayName(): string;
+  abstract get maxFrom(): number;
+  abstract get maxDestinations(): number;
+
+  // Validation helpers
+  public abstract validateAddress(address: string): boolean;
+  public abstract validateValue(value: BigNumber): boolean;
+  public abstract validateAddDestination(address: string, value: BigNumber): boolean;
```
argjv: What is `validateAddDestination` and `validateAddSender`?

src/coin/baseCoin.ts
```diff
@@ -0,0 +1,30 @@
...
+export abstract class BaseCoin {
+  public network: Network;
+  protected staticsCoin: Readonly<StaticsBaseCoin>;
+
+  constructor(network: Network) {
+    this.network = network;
+  }
+
+  // Getters
+  abstract get displayName(): string;
```
argjv: Doc: does it return the full name? ticker? statics name?

src/coin/baseCoin.ts
```diff
@@ -0,0 +1,30 @@
+import { Network, ITransaction, ISignature, IKey, TransactionType } from "..";
+import { BaseCoin as StaticsBaseCoin } from '@bitgo/statics';
+import BigNumber from "bignumber.js";
+
+export abstract class BaseCoin {
```
argjv: Every method in this class should be documented. Some are intuitive (for those who worked on the platform) and some are not.
Please add those docs to serve as guidelines for future devs.

src/coin/baseCoin.ts
```diff
@@ -0,0 +1,30 @@
...
+  public network: Network;
+  protected staticsCoin: Readonly<StaticsBaseCoin>;
+
+  constructor(network: Network) {
+    this.network = network;
+  }
+
+  // Getters
+  abstract get displayName(): string;
+  abstract get maxFrom(): number;
```
cooncesean: What does this represent? Max from what?

src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+export interface IKey {
+  isValid(): boolean;
+}
+
+export enum Network {
+  Test,
+  Main
+}
+
+export enum TransactionType {
```
argjv: For tron and some other we will need `WalletInitialization` or something like that.

src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  NotSet,
```
argjv: What is `NotSet`? I think we should be pretty strict with this enum from the beginning and only add it in the future if we have no choice.

src/coin/baseCoin.ts
cooncesean: And I see that Tron's implementation is `1` ... will that be the default return val for most other coins? If so, should we implement that fcn definition in the base class and have any differing subclasses override it where necessary?

src/transactionBuilder.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.coin = new coinType(network);
+
+    this.fromAddrs = new Array<string>();
+    this.destination = new Array<Destination>();
+    this.signatures = new Array<ISignature>();
+
+    this.transactionType = TransactionType.NotSet;
+  }
+
+  from(rawTransaction: any) {
```
argjv: I just realized this is confusing since we have `addFrom` for the origin of the transaction. We can either rename it or turn it into a constructor `constructor(private coinType: new (network) => T, network: Network, rawTransaction: any) {}`

src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export interface ISignature {
+  isValid(): boolean;
+}
+
+/**
+ * Specifies the members expected for a Transaction
+ */
+export interface ITransaction {
```
cooncesean: What does the `I` prefix stand for?

src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  rawTx: any;
+  parsedTx: any;
+
+  isValid(): boolean;
+}
+
+export interface IKey {
```
cooncesean: Ditto.

src/index.ts
cooncesean: When is `NotSet` ever used? I thought transactions were only one of two types - send and receive?

src/transactionBuilder.ts
```diff
@@ -0,0 +1,102 @@
+import { BaseCoin } from "./coin/baseCoin";
+import { ITransaction, Destination, ISignature, IKey, TransactionType, Network } from ".";
+import BigNumber from "bignumber.js";
+
+
+export default class TransactionBuilder<T extends BaseCoin> {
+  private coin: T;
```
cooncesean: What is `T`?

src/transactionBuilder.ts
```diff
@@ -0,0 +1,102 @@
...
+      throw new Error(`${this.coin.displayName} does not support more than ${this.coin.maxFrom} destinations`);
+    }
+
+    if (!this.coin.validateAddress(address)) {
+      throw new Error(`${address} is not valid for ${this.coin.displayName}`);
+    }
+
+    this.destination.push(new Destination(address, value));
+  }
+}
```
cooncesean: 👍

src/coin/baseCoin.ts
twtaylorbitgo: these are just helpers to validate these. they are mainly for enforcement in transaction builder. at this point could probably be separated into just validate address and value

src/coin/baseCoin.ts
twtaylorbitgo: Ah yeah I'll add some doc to this guy

src/coin/baseCoin.ts
twtaylor: Sure

src/index.ts
twtaylor: good point

src/coin/baseCoin.ts
twtaylor: yes this will be 1. I'm adding comments right now

src/index.ts
twtaylor: NotSet is just an initial state. since we don't set a direction initially we should explicitly flag this

src/transactionBuilder.ts
twtaylor: hm ok I'll add this, I'm not a huge fan of putting it in the constructor because you would need to specify a side of the tx

src/index.ts
twtaylor: interface

https://basarat.gitbooks.io/typescript/docs/styleguide/styleguide.html#interface

I can change it tbh

src/index.ts
twtaylor: changed

src/index.ts
twtaylor: it's used on initialization. we could make it required so you couldn't do anything with a transaction until its set. in fact I've changed it based on that now

src/transactionBuilder.ts
twtaylor: if you look above its a class that extends BaseCoin. this is generics

src/coin/baseCoin.ts
argjv: Yeah, let's just use validate address and value for now.

twtaylor: done

src/coin/baseCoin.ts
cooncesean: I see the docstring (thanks!) ... would it be clearer to call this method `maxSenders()`?

src/coin/baseCoin.ts
```diff
@@ -0,0 +1,67 @@
...
+   * Validate an address. Returns false if the address is invalid
+   * @param address the address
+   * @param addressFormat the address format - this will be handled by the implementing coin as an enum
+   */
+  public abstract validateAddress(address: string, addressFormat?: string): boolean;
+
+  /***
+   * Validates the value corresponding to an amount can be used for this transaction.
+   */
+  public abstract validateValue(value: BigNumber): boolean;
```
cooncesean: Does this method need a return val? If all it is doing is validating the value (the value of the transaction, right?) -- could we have it return safely if the value is valid and raise an exception otherwise?

Nit-picky, sure. Otherwise, if we insist on return true/false depending on the validity of the value being passed , then it might be clear to call this `isValidValue(): boolean`.

Your call if you want to address...

twtaylorbitgo: yeah it's not a bad idea. honestly I would like this function to return different status codes rather than us parse Errors. or we could create error classes. yeah maybe I'll change this in the next iteration

twtaylor: gonna change this in the next PR to keep it moving

PR #5, opened 2019-10-19
TRON-39 - Initial parse and sign for Tron. Included tron protobuf and helpers.

twtaylorbitgo:
Initial check-in for Tron. This includes building, signing. This also includes tronweb.


src/coin/trx/transaction.ts
```diff
@@ -0,0 +1,27 @@
...
+export class Transaction implements BaseTransaction {
+  constructor() {
+    this.failedParse = false;
+  }
+
+  rawTx: any;
+
+  parsedTx: DecodedTransaction;
+
+  transactionType: TransactionType;
```
argjv: This should be in `BaseTransaction `

src/coin/trx/transaction.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  existingSignatures: Array<string>;
+
+  failedParse: boolean;
```
argjv: This should manifest as an exception when the parse or build method is called. Maintaining the state can lead to errors if the user is not aware that he has to call it or devs to maintain it.

src/coin/trx/transaction.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  /**
+   * We have to get this from an external source.
+   */
+  txID: string;
+
+  isValid(): boolean {
```
argjv: DItto. The `isValid()` is implicit when we parse or build a transaction. If it doesn't throw in any of the steps, it is valid.
You could, however, have a static method that takes a raw transaction and checks if it is valid, aka it can parse it: `isValid(rawTransaction: any): boolean`

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+const tronproto = require('../../../resources/trx/protobuf/tron_pb');
+const contractproto = require('../../../resources/trx/protobuf/Contract_pb');
+
+import * as assert from 'assert';
+import { TransferContract, DecodedTransaction, AccountPermissionUpdateContract, ContractType } from './iface';
+
+export type ByteArray = number[];
+export type RawTransaction = { txID: string, signature?: string[] };
+
+export default class TronHelpers {
```
argjv: Can we call it Utils? We may have to export some of these methods. Rename the file too.

src/coin/trx/iface.ts
```diff
@@ -0,0 +1,37 @@
+export interface AccountPermissionUpdateContract {
```
argjv: I can't believe they don't have any types, it's 2019! If we feel generous, one day we can add them to the @types project https://github.com/DefinitelyTyped/DefinitelyTyped#how-can-i-contribute

src/coin/trx/transaction.ts
argjv: Same for `Signature` and `Key`.

src/coin/trx/index.ts
```diff
@@ -1,37 +1,91 @@
...
   
-  public validateValue(value: BigNumber): boolean {
-    // TODO: tbd
-    return true;
+  public validateValue(value: BigNumber) {
+    if (value.isLessThanOrEqualTo(0)) {
+      throw new Error('Value cannot be below zero.');
+    }
+
+    // TODO: is there a max value for Tron sending?
```
argjv: Usually, this is the max number the protocol can handle. I would assume that since the full node is in Java, the max is the max long `9223372036854775807`.
Also, is this method supposed to validate the value in base units (sun)? or TRX?

src/coin/trx/transaction.ts
```diff
@@ -0,0 +1,27 @@
+import { BaseTransaction, TransactionType } from "../..";
+import { DecodedTransaction } from "./iface";
+
+export class Transaction implements BaseTransaction {
+  constructor() {
```
argjv: This should take a network object as param, then use it to display the name and get static values.

src/coin/trx/index.ts
```diff
@@ -55,5 +109,4 @@ export default class Trx extends BaseCoin {
       this.staticsCoin = coins.get('TTRX');
```
argjv: The class should have a network object (or the base coin) as attribute that is passed in the constructor.

src/coin/trx/index.ts
```diff
@@ -1,37 +1,91 @@
...
+    try {
+      signedTx = TronHelpers.signTransaction(privateKey.key, rawTx);
+
+      if (signedTx.signature && signedTx.signature.length > 0) {
+        sig.signature = signedTx.signature[0];
+      } else {
+        throw new Error('Failed to sign transaction.');
+      }
+    } catch (e) {
+      sig.failedSigning = true;
```
argjv: No need to catch it, this should be a hard fail.

src/coin/trx/index.ts
```diff
@@ -1,37 +1,91 @@
...
 
-  public parseTransaction(rawTransaction: any): Transaction {
-    throw new Error("Method not implemented.");
+    // store our transaction
+    tx.rawTx = rawTransaction;
+    tx.transactionType = transactionType;
+
+    // try to parse our transaction
+    let parsedTx: DecodedTransaction;
+    try {
```
argjv: No need for the `try/catch`, we should throw if it fails to decode it.

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+  static generateAccount(): Account {
+    return tronweb.utils.accounts.generateAccount();
+  }
+  
+  static isBase58Address(address: string): boolean {
+    return tronweb.utils.crypto.isAddressValid(address);
+  }
+
+  static isAddress(hexAddress: string): boolean {
```
argjv: s/`isAddress`/`isHexAddress`/g

s/`hexAddress`/`address`/g

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+    return tronweb.isAddress(hexAddress);
+  }
+
+  static hexStr2byteArray(str: string): ByteArray {
```
argjv: `getByteArrayFromHexAddress` ?

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+    return tronweb.utils.code.hexStr2byteArray(str);
+  }
+
+  static byteArray2hexStr(arr: ByteArray): string {
```
argjv: `getHexAddressFromByteArray` ?

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  static signString(message: string, privateKey: string | ByteArray, useTronHeader: boolean = true): string {
+    return tronweb.Trx.signString(message, privateKey, useTronHeader);
+  }
+
+  static computeAddress(pubBytes: number[] | string): ByteArray {
+    return tronweb.utils.crypto.computeAddress(pubBytes);
+  }
+
+  static getHexFromBase58Address(base58: string): string {
```
argjv: s/`getHexFromBase58Address `/`getHexAddressFromBase58Address`/g

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  static signTransaction(privateKey: string | ByteArray, transaction: RawTransaction): RawTransaction {
+    return tronweb.utils.crypto.signTransaction(privateKey, transaction);
+  }
+
+  static signString(message: string, privateKey: string | ByteArray, useTronHeader: boolean = true): string {
+    return tronweb.Trx.signString(message, privateKey, useTronHeader);
+  }
+
+  static computeAddress(pubBytes: number[] | string): ByteArray {
```
argjv: `getRawAddressFromPubKey` ?

src/coin/trx/helpers.ts
```diff
@@ -0,0 +1,176 @@
...
+      timestamp: raw.timestamp,
+      contracts: raw.contracts,
+    };
+  }
+
+  /**
+   * Indicates whether the passed string is a safe hex string for tron's purposes.
+   * @param str A valid hex string must be a string made of numbers and characters and has an even length.
+   */
+  static safeHexString(str: string): Boolean {
```
argjv: Since it returns a boolean, it should be named `isValidHex()`

src/coin/trx/transaction.ts
twtaylorbitgo: oh yup, it is, I overloaded it here for some reason

twtaylorbitgo: I've converted BaseTransaction to a class

src/coin/trx/transaction.ts
twtaylorbitgo: I think I found a better solution for this. You'll find an answer in the next commit. Basically the idea is to surface what kind of error (rather than just Error) and that it bubbles up to a caller properly, to be handled.

I've added an errors class to solve this and removed most of the isValid comments.

src/coin/trx/helpers.ts
twtaylorbitgo: sure

twtaylorbitgo: I want to put these under a namespace then too or module

src/coin/trx/iface.ts
twtaylorbitgo: yeah. there's a half-baked one out there submitted to definitely typed but it's not useful at all

src/coin/trx/index.ts
twtaylorbitgo: I think we want to have it be TRX instead of Sun. We have bignumber so might as well use it.

src/coin/trx/transaction.ts
twtaylorbitgo: sure

src/coin/trx/index.ts
twtaylorbitgo: ah it does. I've refernced the baseCoin as the network. We'll still have to say what coin to get for statics

twtaylorbitgo: we might be able to make this a Map like BitGoJS does

src/coin/trx/helpers.ts
twtaylorbitgo: oh nice

src/coin/trx/helpers.ts
twtaylorbitgo: nice catch

src/coin/trx/helpers.ts
twtaylorbitgo: done

src/coin/trx/helpers.ts
twtaylorbitgo: done

src/coin/trx/helpers.ts
twtaylorbitgo: how about getByteArrayFromPubKey?

twtaylorbitgo: actually that doesn't make any sense either - yeah let's call ByteArray addresses raw, hex addresses hex and base58 base58

src/coin/trx/helpers.ts
twtaylorbitgo: done

twtaylorbitgo:
I've gotten the base tests to pass. I can add more Tron-specific in the next PR.

src/coin/trx/index.ts
```diff
@@ -1,36 +1,106 @@
...
+      parsedTx = Utils.decodeTransaction(rawTransaction);
+    } catch (e) {
+      throw new ParseTransactionError('Failed to decode transaction.');
+    }
+
+    switch (parsedTx.contractType) {
```
argjv: Is this gonna work for init wallet transactions?

twtaylorbitgo: hm ok fixing this up

src/coin/trx/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+    return {
+      contractType,
+      contract,
+      expiration: rawTransaction.expiration,
+      timestamp: rawTransaction.timestamp,
+    };
+  }
+
+  /**
+   * Decodes a transaction's raw field from a base64 encoded string. This is a protobuf representation.
```
glowang: Just need to clarify - the comment implies that we are returning the crude/raw protobuf representation of the result. But in reality we call toObject() on it, so it is not exactly a protobuf representation. However, we do return the object that is typed "raw" in the tron.Proto file.

src/coin/trx/transaction.ts
```diff
@@ -0,0 +1,21 @@
+import { DecodedTransaction } from "./iface";
+import { BaseTransaction } from "../baseCoin/iface";
+import { Network } from "../baseCoin/enum";
+
+export class Transaction extends BaseTransaction {
+  rawTx: any;
+
+  parsedTx: DecodedTransaction;
+
+  existingSignatures: Array<string>;
```
argjv: What do we need this one for? Aren't the signatures inside the `parsedTx`?

twtaylorbitgo: they're not, parsedTx actually doesn't contain any of that information (it's more a representation of the raw object) which is here:

https://github.com/BitGo/bitgo-account-lib/blob/5f282588701778a4421c75fa61f42713f56e95b9/resources/trx/protobuf/tron.proto#L319

what we want is outside, which is here:

https://github.com/BitGo/bitgo-account-lib/blob/5f282588701778a4421c75fa61f42713f56e95b9/resources/trx/protobuf/tron.proto#L337

src/coin/trx/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+  static getBase58AddressFromByteArray(address: ByteArray): string {
+    return tronweb.utils.crypto.getBase58CheckAddress(address);
+  }
+
+  static getBase58AddressFromHex(hex: string): string {
+    const arr = Utils.getByteArrayFromHexAddress(hex);
+    return Utils.getBase58AddressFromByteArray(arr);
+  }
+
+  static signTransaction(privateKey: string | ByteArray, transaction: RawTransaction): RawTransaction {
```
glowang: This return type "RawTransaction" does not match with the SDK basecoin implementation 
https://github.com/BitGo/BitGoJS/blob/05d470e780ef679fa8779aa4143d562d069e686a/modules/core/src/v2/baseCoin.ts#L549 
  abstract signTransaction(params: SignTransactionOptions): SignedTransaction;

glowang: I think we need to agree on one return type here

twtaylorbitgo: it's not supposed to

src/coin/trx/utils.ts
```diff
@@ -0,0 +1,179 @@
...
+import { TransferContract, DecodedTransaction, AccountPermissionUpdateContract, ContractType } from './iface';
+
+export type ByteArray = number[];
+export type RawTransaction = { txID: string, signature?: string[] };
+
+/**
+ * Tron-specific helpers
+ */
+export default class Utils {
+  static generateAccount(): Account {
```
glowang: type Account should have been imported from ./iface too. It is throwing in the SDK as unrecognized type.

src/coin/trx/utils.ts
twtaylorbitgo: I'll clarify the comment but I'm only talking about the parameter here, it's serialized as base64 so that you can use Buffer.from() on it and be able to convert to the protobuf representation easier

glowang: Gotcha. Yea, it makes sense for the parameter.

src/coin/trx/utils.ts
twtaylorbitgo: yeah that's great, that works

src/coin/trx/index.ts
twtaylorbitgo: not done yet

src/coin/trx/utils.ts
```diff
@@ -0,0 +1,180 @@
...
+    return false;
+  }
+
+  return true;
+}
+
+/** Deserialize the portion of the txHex which corresponds with the details of the transfer
+ * @param transferHex is the value property of the "parameter" field of contractList[0]
+ * */
+export function decodeTransferContract(transferHex: string): TransferContract {
```
marktoda: Do you think we can make this a coin-generic "decodeTxHex" function? This would help avoid things like:
```
if (coin is trx) {
  const deserializedTx = accountLib.trx.util.decodeTransferContract(txHex);
} else if (coin is x) {
  const deserializedTx = accountLib.x.util.decodeXBlah(txHex);
}
```

twtaylorbitgo: yeah this might be a case for generic utils, we don't have a pattern like that in the spec. definitely something to think about. (this is a tron-specific utility file)

PR #7, opened 2019-10-22
TRON-39: decode accountPermissionUpdateContract Tx

glowang:
* We need this function to deserialize the special type of tx that
updates user's permission


src/coin/trx/utils.ts
```diff
@@ -174,7 +181,34 @@ export function decodeTransferContract(transferHex: string): TransferContract {
 }
 
 export function decodeAccountPermissionUpdateContract(base64: string): AccountPermissionUpdateContract {
```
marktoda: jsdoc

src/coin/trx/utils.ts
```diff
@@ -174,7 +181,34 @@ export function decodeTransferContract(transferHex: string): TransferContract {
...
-  throw new Error();
-  // TODO: build
+  const accountUpdateContract = contractproto.AccountPermissionUpdateContract.deserializeBinary(Buffer.from(base64, 'base64')).toObject();
```
marktoda: The whole first part of `decodeTransferContract` (line 156-165) should be essentially the same as here. I think we should:
- Name the parameters the same (and accurately) --> contractBase64 or something seems accurate
- dedupe the try/catch and deserialization
- dedupe getBase58AddressFromByteArray(getByteArrayFromHexAddress(Buffer.from(accountUpdateContract.ownerAddress, 'base64').toString('hex')));

src/coin/trx/utils.ts
```diff
@@ -174,7 +181,34 @@ export function decodeTransferContract(transferHex: string): TransferContract {
...
+  assert(accountUpdateContract.ownerAddress);
+  assert(accountUpdateContract.owner);
+  assert(accountUpdateContract.hasOwnProperty('activesList'));
+
+  const ownerAddress = getBase58AddressFromByteArray(getByteArrayFromHexAddress(Buffer.from(accountUpdateContract.ownerAddress, 'base64').toString('hex')));
+  const owner: Permission = {
+    type: accountUpdateContract.owner.permissionName,
+    threshold: accountUpdateContract.owner.threshold,
```
marktoda: This whole thing is duplicated a few other times, could be extracted to a function

marktoda:
Also please make sure the tests pass (they are failing for me locally but I may be doing something wrong)

glowang:
> Also please make sure the tests pass (they are failing for me locally but I may be doing something wrong)

Some other parts of the lib still have compilation issues.

src/coin/trx/utils.ts
```diff
@@ -174,7 +181,34 @@ export function decodeTransferContract(transferHex: string): TransferContract {
 }
 
 export function decodeAccountPermissionUpdateContract(base64: string): AccountPermissionUpdateContract {
-  throw new Error();
-  // TODO: build
+  const accountUpdateContract = contractproto.AccountPermissionUpdateContract.deserializeBinary(Buffer.from(base64, 'base64')).toObject();
+  assert(accountUpdateContract.ownerAddress);
```
twtaylor: these will be changed in my PR

src/coin/trx/utils.ts
twtaylorbitgo: ok this is actually fixed in my PR, so I don't know if she needs to fix this right now

PR #9, opened 2019-10-24
Few improvements and tests

argjv:


src/coin/baseCoin/iface.ts
```diff
@@ -1,25 +1,18 @@
...
-}
-
 export interface BaseKey {
   key: any;
-}
\ No newline at end of file
+}
+
+export interface Destination {
+  address: any,
```
twtaylorbitgo: should make this a BaseAddress, so it's extensible

src/coin/baseCoin/index.ts
```diff
@@ -1,19 +1,22 @@
...
    * Our coin from statics
    */
   protected staticsCoin: Readonly<StaticsBaseCoin>;
+  protected _network: AccountNetwork;
 
   /**
    * Constructor
    * @param network the network for this coin
    */
-  constructor(public network: NetworkType) { }
```
twtaylorbitgo: note that you can change this to protected network: NetworkType, it will automatically make it a protected variable attached to this class

src/transaction.ts
```diff
@@ -0,0 +1,36 @@
+import { BaseAddress, Destination } from "./coin/baseCoin/iface";
+import { AccountNetwork } from "@bitgo/statics";
+import { TransactionType } from "./coin/baseCoin/enum";
+
+/**
+ * Specifies the members expected for a Transaction
+ */
+export abstract class BaseTransaction {
+  private _network: AccountNetwork;
+  protected _id: string;
```
twtaylorbitgo: need a comment about what this is intended to be for any implementers. txId?

src/coin/trx/network/base.ts
```diff
@@ -77,52 +58,48 @@ export class TrxBase extends BaseCoin {
...
    * Parse transaction takes in raw JSON directly from the node.
+   * @param rawTransaction The Tron transaction in JSON format.
    */
-  public parseTransaction(rawTransaction: any, transactionType: TransactionType): Transaction {
-    const tx = new Transaction(this.network);
-
-    if (typeof rawTransaction !== 'string') {
-      throw new ParseTransactionError('Raw transaction needs to be a JSON encoded string.');
+  public parseTransaction(rawTransaction: TransactionReceipt | string): Transaction {
+    if (typeof rawTransaction === 'string') {
```
twtaylorbitgo: probably needs a comment about the two ways we can parse this

src/transaction.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  get destinations(): Destination[] {
+    return this._destination;
+  }
+
+  get senders(): BaseAddress[] {
+    return this._fromAddresses;
+  }
+
+  abstract toJson(): any;
```
twtaylorbitgo: i love this pattern

test/unit/transactionBuilder.ts
```diff
@@ -1,41 +1,25 @@
...
+    txBuilder.from(null);
     txBuilder.sign({ key: ''  }, { address: 'fakeAddress' });
   });
 
   it('should build an existing transaction that is valid', () => {
-    const validTx = { isValid: () => true, finalTx: null, rawTx: null, tx: null, transactionType: TransactionType.Send };
-    coin.setVariable({ buildTransaction: validTx, validateAddress: true });
-
-    txBuilder.from(null, TransactionType.Send);
+    txBuilder.from(null);
```
twtaylorbitgo: we talked about this, I think we can defer a spec type test until a later commit or based on a sinon implementation

src/coin/trx/network/base.ts
argjv: Added in the method comment

PR #14, opened 2019-10-25
TRON-82 - Remove address param from builder.sign()

argjv:
The address in the signing step is redundant. At the signing step, we should already have all the information needed in the transaction to verify it is valid.
This check can easily be replaced by deriving the public key and subsequent addresses from the private key anyway.


twtaylorbitgo:
> LGTM

Who is this guy? :D

PR #15, opened 2019-10-25
TRON-39 - Adding verify signature and function to utils

twtaylorbitgo:


test/unit/coin/trx/util.ts
```diff
@@ -84,6 +84,11 @@ describe('Util library should', function() {
     should.deepEqual(bytes, addrBytes);
   });
 
+  it('should verify a signed message', () => {
+    const hexEncodedMessage = Buffer.from(txt).toString('hex');
+    should.ok(Utils.verifySignature(hexEncodedMessage, base58, signedString, true));
+  });
+
```
argjv:
```suggestion
it('should fail to verify a signed message if the message is not in hex', () => {
    should.throws(() => Utils.verifySignature(txt, base58, signedString, true));
  });

  it('should fail to verify a signed message if the address is not in base58', () => {
    const hexEncodedString = Buffer.from(txt).toString('hex');
    should.throws(() => Utils.verifySignature(hexEncodedString, addressHex, signedString, true));
  });

  it('should fail to verify a signed message if the signature is not in hex', () => {
    const hexEncodedString = Buffer.from(txt).toString('hex');
    should.throws(() => Utils.verifySignature(hexEncodedString, base58, 'abc', true));
  });
```

twtaylor: added

PR #16, opened 2019-10-25
[TRON-81] Add utility to future date Tron transactions

marktoda:
The function deserializes tx hex, updates the expiration date, and
reserializes the tx hex.

I was able to send a successful extended multisig transaction using this code in coins-sandbox:
```
   ... normal send stuff

    const future_date_amount_ms = 4000;
    tx.raw_data.expiration = tx.raw_data.expiration += future_date_amount_ms;
    const updatedTxHex = bitgoAccountLib.Trx.Utils.extendTransaction(tx.raw_data_hex, future_date_amount_ms);
    tx.raw_data_hex = updatedTxHex; 
    tx.txID = calculateTxid(tx);

   ... signature, signature, send

```

### DIFF
--- a/src/coin/trx/transaction.ts
+++ b/src/coin/trx/transaction.ts
@@ -1,12 +1,14 @@
-import {RawData, TransactionReceipt, TransferContract} from "./iface";
+import { protocol } from '../../../resources/trx/protobuf/tron';
+const crypto = require('crypto');
+import { RawData, TransactionReceipt, TransferContract } from "./iface";
 import { BaseCoin as CoinConfig } from "@bitgo/statics";
 import { BaseTransaction } from "../../transaction";
 import { decodeTransaction } from "./utils";
 import { ContractType} from "./enum";
 import BigNumber from "bignumber.js";
-import { ParseTransactionError } from "../baseCoin/errors";
+import { ParseTransactionError, ExtendTransactionError } from "../baseCoin/errors";
 import { TransactionType } from "../baseCoin/";
-import {BaseKey} from "../baseCoin/iface";
+import { BaseKey } from "../baseCoin/iface";
 
 export class Transaction extends BaseTransaction {
   private _decodedRawDataHex: RawData;
@@ -21,13 +23,6 @@ export class Transaction extends BaseTransaction {
       this._id = rawTransaction.txID;
       this._transaction = rawTransaction;
       this._decodedRawDataHex = decodeTransaction(rawTransaction.raw_data_hex);
-      const senderAddress = {
-        address: (this._decodedRawDataHex.contract[0] as TransferContract).parameter.value.owner_address
-      };
-      this._fromAddresses = [senderAddress];
-
-      this._validFrom = Number(this._decodedRawDataHex.timestamp);
-      this._validTo = Number(this._decodedRawDataHex.expiration);
 
       // Destination depends on the contract type
       this.recordRawDataFields(this._decodedRawDataHex);
@@ -39,6 +34,16 @@ export class Transaction extends BaseTransaction {
    * @param rawData Object from a tron transaction
    */
   private recordRawDataFields(rawData: RawData) {
+    // Contract-agnostic fields
+    const senderAddress = {
+      address: (rawData.contract[0] as TransferContract).parameter.value.owner_address
+    };
+    this._fromAddresses = [senderAddress];
+
+    this._validFrom = rawData.timestamp;
+    this._validTo = rawData.expiration;
+
+    // Contract-specific fields
     switch (rawData.contractType) {
       case ContractType.Transfer:
         this._type = TransactionType.Send;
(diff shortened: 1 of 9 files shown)


src/coin/trx/utils.ts
```diff
@@ -121,6 +129,26 @@ export function decodeTransaction(hexString: string): RawData {
...
 }
 
+/**
+ * Extends the expiration date of a transaction. This must be done before any signing because the txHex and txId
+ * are affected by the new expiration date
+ * @param hexString raw_data_hex field from tron transactions
+ * @param extensionMs The amount of time to extend expiration by in milliseconds
+ * @returns Hex string with the new expiration date
+ */
+export function extendTransaction(hexString: string, extensionMs: number): string {
```
twtaylorbitgo: if this is a raw_data_hex field, it should be extending off that class, also extensionMs should be an interface over a set of params

src/coin/trx/utils.ts
```diff
@@ -218,6 +246,18 @@ export function decodeAccountPermissionUpdateContract(base64: string): AccountPe
...
+/**
+ * Calculate the txid of this transaction based on its tx hex string (i.e. raw_data_hex)
+ * The txid needs to be recalculated any time the hex is updated, for instance when the transaction
+ * expiration date is extended.
+ * @param txHexString The raw_data_hex to calculate txid for
+ * @returns transaction id for this hex.
+ */
+export function calculateTransactionId(txHexString: string): string {
+  const txBuffer = Buffer.from(txHexString, 'hex');
+  return crypto.createHash('sha256').update(txBuffer).digest('hex');
```
twtaylorbitgo: again, actually, this should go into when we build a transaction. this is good stuff though. this all should be added to build/sign/from flow pretty easily

src/coin/trx/utils.ts
marktoda: i dont know what ou mean

src/coin/trx/utils.ts
marktoda: also dont know what you mean. I am using this function in platform buildTransaction function

marktoda:
https://github.com/BitGo/platform/pull/1445/files <-- corollary platform diff

src/coin/trx/utils.ts
twtaylor: so again, I think in general instead of adding these functions, we should decorate what going in-between the from and sign step. so when we call from() we add the tx details. then we can modify the expiration attribute. this will then auto-recompute the txId cause a change was made before we call sign

argjv: This should be split into 3 places, but not in the utils:
1. [src/coin/trx/transaction](https://github.com/BitGo/bitgo-account-lib/blob/master/src/coin/trx/transaction.ts): where all this logic goes changing the Transaction internal status.
2. [src/coin/trx/network/coin](https://github.com/BitGo/bitgo-account-lib/blob/master/src/coin/trx/network/coin.ts): add a method that will be called by the Transaction Builder that basically calls the method in 1) from the transaction in memory. 
3. [src/transactionBuilder](https://github.com/BitGo/bitgo-account-lib/blob/master/src/transactionBuilder.ts): add a method (step) like `extendValidTo(extension: Milliseconds)` or something like that.

src/coin/trx/utils.ts
argjv: This should be called in the build step of the Tron Transaction. Again, TransactionBuilder -> TrxBase -> Transaction

test/unit/coin/trx/trx.ts
```diff
@@ -114,4 +114,36 @@ describe('Tron test network', function() {
...
+      should.throws(() => txBuilder.extendValidTo(10000));
+    });
+
+    it('should extend an unsigned tx', () => {
+      const extendMs = 10000;
+      txBuilder.from(JSON.parse(JSON.stringify(UnsignedBuildTransaction)));
+      txBuilder.extendValidTo(extendMs);
+      const tx = txBuilder.build();
+
+      tx.id.should.not.equal(UnsignedBuildTransaction.txID);
```
argjv: Also, check for the new id. It's ok to hardcode it since it should always be the same.

src/coin/trx/transaction.ts
```diff
@@ -39,6 +34,16 @@ export class Transaction extends BaseTransaction {
    * @param rawData Object from a tron transaction
    */
   private recordRawDataFields(rawData: RawData) {
+    // Contract-agnostic fields
+    const senderAddress = {
+      address: (this._decodedRawDataHex.contract[0] as TransferContract).parameter.value.owner_address
```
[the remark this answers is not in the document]
marktoda: done

src/transactionBuilder.ts
```diff
@@ -56,4 +56,12 @@ export class TransactionBuilder {
     this.transaction = this.coin.buildTransaction(this.transaction);
     return this.transaction;
   }
+
+  /**
+   * Extend the validity of this transaction by the given amount of time
+   * @param extensionMs The number of milliseconds to extend the validTo time
+   */
+  extendValidTo(extensionMs: number) {
```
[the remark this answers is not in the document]
twtaylorbitgo: transactionBuilder is intended to be used generally across all coins, the more extensible function would be like extend(any) in general. however this is fine for now until we implement a coin that doesn't need this function (and thus would cause a no-op)

PR #17, opened 2019-10-28
Try protobufjs to generate the protobuff code

argjv:
Change the tool to generate the static code from the proto files. The old one generated code using eval under the hood which is blocked by our client app.

The new code is static and doesn't use reflection. In addition, it has types!


twtaylorbitgo:
Can you include an .md file with the generation technique

PR #18, opened 2019-10-28
Update package.json

argjv:
Update package.json with new commands to generate the protobuf js libs, and clean up dependencies


argjv:
Good point. Maybe this is possible with webpack. Not sure how though.

To that point, we are thinking about making this lib plugin-based, in which each coin implementation will be a separate project that gets imported by the user depending on what coins they want to support.

PR #21, opened 2019-10-30
Changes protobuf libs to use protobufjs typings instead

twtaylorbitgo:
Tries to match the case of what comes in from the node and exits it exactly. Also introduces the same typings as protobufjs's TS implementation


twtaylorbitgo:
This will need to be re-based over Toda's changes, once they're made

PR #26, opened 2019-11-07
TRON-133 use bitcoin key generation libs

argjv:
[TRON-133](https://bitgoinc.atlassian.net/browse/TRON-133) Use bitcoin account generation methods instead of Tron Web ones.

### DIFF
--- a/src/coin/trx/utils.ts
+++ b/src/coin/trx/utils.ts
@@ -1,6 +1,7 @@
 const crypto = require('crypto');
 const tronweb = require('tronweb');
 import { protocol } from '../../../resources/trx/protobuf/tron';
+import { HDNode, ECPair } from 'bitgo-utxo-lib';
 
 import * as assert from 'assert';
 import {
@@ -20,8 +21,38 @@ import { UtilsError } from '../baseCoin/errors';
 export type TronBinaryLike = ByteArray | Buffer | Uint8Array | string;
 export type ByteArray = number[];
 
-export function generateAccount(): Account {
-  return tronweb.utils.accounts.generateAccount();
+/**
+ * Generate a Tron account offline using known bitcoin libraries
+ * @param seed Optional random seed
+ * @return {Account}
+ */
+export function generateAccount(seed?: Buffer): Account {
+  if (!seed) {
+    seed = crypto.randomBytes(512 / 8);
+  }
+
+  const extendedKey = HDNode.fromSeedBuffer(seed);
+
+  // convert our prv, pub
+  const prv = extendedKey.keyPair.getPrivateKeyBuffer().toString('hex').toUpperCase();
+
+  const keyPair = ECPair.fromPrivateKeyBuffer(Buffer.from(prv, 'hex'));
+
+  const priKeyBytes = getByteArrayFromHexAddress(keyPair.getPrivateKeyBuffer().toString('hex'));
+  const pubKeyBytes = getPubKeyFromPriKey(priKeyBytes);
+  const publicKey = getHexAddressFromByteArray(pubKeyBytes);
+
+  // used for meaningful address conversion
+  const addressBytes = getAddressFromPriKey(priKeyBytes);
+
+  return {
+    privateKey: prv,
+    publicKey,
+    address: {
+      base58: getBase58AddressFromByteArray(addressBytes),
+      hex: getHexAddressFromByteArray(addressBytes),
+    },
+  };
 }
 
 export function isBase58Address(address: string): boolean {
--- a/test/unit/coin/trx/util.ts
+++ b/test/unit/coin/trx/util.ts
@@ -1,3 +1,4 @@
+const EC = require('elliptic').ec;
 import { TransferContract, AccountPermissionUpdateContract } from '../../../../src/coin/trx/iface';
 import { Utils } from '../../../../src/coin/trx/index';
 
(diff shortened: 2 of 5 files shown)


test/unit/coin/trx/util.ts
```diff
@@ -16,7 +16,16 @@ describe('Util library should', function() {
   const pub = '046EBFB90C396B4A3B992B727CB4714A32E2A6DE43FDB3EC266286AC2246D8FD1E23E12C0DEB752C631A9011BBF8B56E2FBAA20E99D3952F0A558D11F96E7C1C5D';
   const addressHex = '412C2BA4A9FF6C53207DC5B686BFECF75EA7B80577';
   const base58 = 'TDzm1tCXM2YS1PDa3GoXSvxdy4AgwVbBPE';
-  const addrBytes = [ 65, 44, 43, 164, 169, 255, 108, 83, 32, 125, 197, 182,134,191,236,247,94,167,184,5,119 ]
+  const addrBytes = [ 65, 44, 43, 164, 169, 255, 108, 83, 32, 125, 197, 182,134,191,236,247,94,167,184,5,119 ];
+
+  it('generate a valid account', () => {
+    const seed = Buffer.alloc(32);
```
twtaylorbitgo: we should pass a known seed to this in the test

argjv: It is a known seed, it's all zeros

twtaylorbitgo: eh fair enough, I think we should be a little more explicit but that works

test/unit/coin/trx/util.ts
```diff
@@ -16,7 +16,16 @@ describe('Util library should', function() {
   const pub = '046EBFB90C396B4A3B992B727CB4714A32E2A6DE43FDB3EC266286AC2246D8FD1E23E12C0DEB752C631A9011BBF8B56E2FBAA20E99D3952F0A558D11F96E7C1C5D';
   const addressHex = '412C2BA4A9FF6C53207DC5B686BFECF75EA7B80577';
   const base58 = 'TDzm1tCXM2YS1PDa3GoXSvxdy4AgwVbBPE';
-  const addrBytes = [ 65, 44, 43, 164, 169, 255, 108, 83, 32, 125, 197, 182,134,191,236,247,94,167,184,5,119 ]
+  const addrBytes = [ 65, 44, 43, 164, 169, 255, 108, 83, 32, 125, 197, 182,134,191,236,247,94,167,184,5,119 ];
+
+  it('generate a valid account', () => {
```
twtaylorbitgo: we should have an additional test where we pass no see (to make sure the crypto.randomBytes path spits out a valid key

src/coin/trx/utils.ts
```diff
@@ -20,8 +22,66 @@ import { UtilsError } from '../baseCoin/errors';
...
+    seed = crypto.randomBytes(512 / 8);
+  }
+
+  const extendedKey = HDNode.fromSeedBuffer(seed);
+
+  // convert our prv, pub
+  const prv = extendedKey.keyPair.getPrivateKeyBuffer().toString('hex').toUpperCase();
+  const pub = extendedKey.keyPair.getPublicKeyBuffer().toString('hex').toUpperCase();
+
+  validateKeyPair(pub, prv);
```
tylerlevine: Why do we need to validate here? Does HDNode give you invalid keys with some nonzero probability?

src/coin/trx/utils.ts
```diff
@@ -20,8 +22,66 @@ import { UtilsError } from '../baseCoin/errors';
...
+    publicKey,
+    address: {
+      base58: getBase58AddressFromByteArray(addressBytes),
+      hex: getHexAddressFromByteArray(addressBytes),
+    },
+  };
+}
+
+/**
+ * Validate the public and private key are valid points of the secp256k1 elliptic curve.
```
tylerlevine: Nit: The private key is a scalar (number), only the public key is a point on secp.

src/coin/trx/utils.ts
```diff
@@ -20,8 +22,66 @@ import { UtilsError } from '../baseCoin/errors';
...
+
+  // convert our prv, pub
+  const prv = extendedKey.keyPair.getPrivateKeyBuffer().toString('hex').toUpperCase();
+  const pub = extendedKey.keyPair.getPublicKeyBuffer().toString('hex').toUpperCase();
+
+  validateKeyPair(pub, prv);
+
+  // 0 pads the private key
+  let priKeyHex = prv;
+  while (priKeyHex.length < 64) {
```
tylerlevine: This has a bit of code smell to it as well, but I don't think it's worth bringing in lodash for one util function.

twtaylorbitgo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

twtaylorbitgo: we'd have to upgrade the webpack targets though

src/coin/trx/utils.ts
argjv: Not really, but this was added in the beginning when we implemented our key generation logic. We wanted to triple check we did it right.

src/coin/trx/utils.ts
argjv: Fixed by using KeyPair objects from bitgo-utxo-lib which automatically pads the prv.

src/coin/trx/utils.ts
tylerlevine: This is better off as a unit test then - we shouldn't be testing our deps in the hot path

PR #41, opened 2019-12-11
Use inputs and outputs nomenclature

argjv:
Destination and sender is not a good name. Inputs and outputs are easier to understand. It also makes it easier to consume by library users.

### DIFF
--- a/src/coin/baseCoin/iface.ts
+++ b/src/coin/baseCoin/iface.ts
@@ -1,7 +1,7 @@
 import BigNumber from 'bignumber.js';
 
 export interface BaseAddress {
-  address: any;
+  address: string;
 }
 
 export interface BaseKey {
--- a/src/coin/baseCoin/iface.ts
+++ b/src/coin/baseCoin/iface.ts
@@ -24,6 +24,6 @@ export type ExtendedKeys = {
   xpub: string;
 }
 
-export interface Destination extends BaseAddress {
+export interface Entry extends BaseAddress {
   value: BigNumber;
 }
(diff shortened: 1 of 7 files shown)


src/coin/baseCoin/iface.ts
```diff
@@ -24,6 +24,6 @@ export type ExtendedKeys = {
   xpub: string;
 }
 
-export interface Destination extends BaseAddress {
+export interface Entry extends BaseAddress {
```
CapnMigraine: nit: unrelated to this change, but can we make `address` in `BaseAddress` a string?

src/coin/trx/transaction.ts
```diff
@@ -48,33 +48,37 @@ export class Transaction extends BaseTransaction {
...
+        const value = new BigNumber((rawData.contract[0] as TransferContract).parameter.value.amount);
+        output = {
           address: (rawData.contract[0] as TransferContract).parameter.value.to_address,
-          value: new BigNumber((rawData.contract[0] as TransferContract).parameter.value.amount),
+          value,
         };
-        senderAddress = {
+        input = {
           address: (rawData.contract[0] as TransferContract).parameter.value.owner_address,
+          value,
```
CapnMigraine: shouldn't we negate input values as we do in entries created by indexer and platform?

src/coin/trx/transaction.ts
```diff
@@ -130,6 +134,16 @@ export class Transaction extends BaseTransaction {
     }
   }
 
+  get signature(): string[] {
```
CapnMigraine: nit: add docstring

src/coin/trx/transaction.ts
argjv: Good point, but these are inputs and outputs that go into separate fields. If they were mixed up, then yeah, we should negate the values.

PR #43, opened 2019-12-27
Add sign and verify message functions

argjv:
[XTZ-29](https://bitgoinc.atlassian.net/browse/XTZ-29) Add methods to sign and verify random string messages using the current Key Pair.

This is used to prove that a sender has control over the private keys.

### DIFF
--- a/src/coin/trx/keyPair.ts
+++ b/src/coin/trx/keyPair.ts
@@ -123,4 +123,32 @@ export class KeyPair {
     }
     throw new Error('Unsupported address format');
   }
+
+  /**
+   * Generates a signature for an arbitrary string with the current private key using keccak256
+   * hashing algorithm. Throws if there is no private key.
+   * @param {string} message to produce a signature for
+   * @return The signature as a buffer
+   */
+  signMessage(message: string): Buffer {
+    const messageToSign = Buffer.from(message).toString('hex');
+    const { prv } = this.getKeys();
+    if (!prv) {
+      throw new Error('Missing private key');
+    }
+    const signature = Utils.signString(messageToSign, prv, true).replace(/^0x/, '');
+    return Buffer.from(signature, 'hex');
+  }
+
+  /**
+   * Verifies a message signature using the current public key.
+   * @param {string} message signed
+   * @param {Buffer} signature to verify
+   * @return True if the message was signed with the current key pair
+   */
+  verifySignature(message: string, signature: Buffer): boolean {
+    const messageToVerify = Buffer.from(message).toString('hex');
+    const address = this.getAddress(AddressFormat.base58);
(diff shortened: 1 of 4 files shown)


src/coin/trx/keyPair.ts
```diff
@@ -123,4 +123,31 @@ export class KeyPair {
...
   }
+
+  /**
+   * Signs a message with the current private key. Throws if there are no private keys.
+   * @param {string} message to produce a signature for
+   * @return The signature as a buffer
+   */
+  signMessage(message: string): Buffer {
+    const messageToSign = Buffer.from(message).toString('hex');
+    const { prv } = this.getKeys();
```
twtaylorbitgo: I would actually pass this to the function, since this implies any message will use the same set of keys or we'll be able to deconstruct them, would this work for anything where we don't have the keys natively?

src/coin/trx/keyPair.ts
```diff
@@ -123,4 +123,31 @@ export class KeyPair {
...
     throw new Error('Unsupported address format');
   }
+
+  /**
+   * Signs a message with the current private key. Throws if there are no private keys.
+   * @param {string} message to produce a signature for
+   * @return The signature as a buffer
+   */
+  signMessage(message: string): Buffer {
+    const messageToSign = Buffer.from(message).toString('hex');
```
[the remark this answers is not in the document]
argjv: Transactions signatures are generated using the transaction id, not the who transaction, so if you put a whole tx here it would produce an invalid signature.

If you pass a transaction id as a message, it will still produce an invalid signature too because:
1) Transactions signatures are made using double SHA-256 while message signatures are generated using keccak256.
2) The `signString()` method always adds a header (TRON or ETH header) to the message.

src/coin/trx/keyPair.ts
argjv: This is a class method, not static. You have to initialize the object with a seed or private key in order to use it.
If you don't have the keys you cannot sign anyway, you need the private key.

The idea of adding these methods here is to take advantage of all the validations done for the seed, prvs, and pubs during initialization and so that users don't have to worry about what format of the private key they need to use.

```
const keyPair = new accountLib.Trx.KeyPair({ prv });
return keyPair.signMessage(message);
```

With these 2 lines, you are running all the validations for the private key, the message, selecting the right format of the private key and generating a valid signature compatible with the underlying protocol.

src/coin/trx/keyPair.ts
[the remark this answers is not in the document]
twtaylorbitgo: Ahhh ok. This looks good then!

src/coin/trx/keyPair.ts
twtaylorbitgo: ah ok, yes, this is sound if we ensure we only contain one key in this class.

PR #44, opened 2020-01-23
Validate transaction implementation

### LINKED ISSUE #37: Tron TransactionBuilder - Implement validateTransaction
# Overview
The Tron `TransactionBuilder` class defines a method called [`validateTransaction()`](https://github.com/BitGo/bitgo-account-lib/blob/master/src/coin/trx/transactionBuilder.ts#L120-L122). The method currently performs zero validation on the raw transaction passed into it.

Let's implement the method to perform the necessary validation. Specifically: verify the transaction has the right transaction id.

# Definition of Done
- Implement Tron's `TransactionBuilder::validateTransaction()` method.
- Add test(s) to cover these changes.

wesgraham:
Fixes #37 

As specified, `validateTransaction()` verifies a Tron transaction has the right transaction id, per the [Tron spec](https://github.com/tronprotocol/documentation/blob/master/TRX/Tron-overview.md). Introduced the crypto library from standard lib.

Unit tests included.

### DIFF
--- a/src/coin/baseCoin/errors.ts
+++ b/src/coin/baseCoin/errors.ts
@@ -43,3 +43,9 @@ export class ExtendTransactionError extends ExtendableError {
     super(message);
   }
 }
+
+export class InvalidIDError extends ExtendableError {
+  constructor(message: string) {
+    super(message);
+  }
+}
--- a/src/coin/baseCoin/errors.ts
+++ b/src/coin/baseCoin/errors.ts
@@ -44,7 +44,7 @@ export class ExtendTransactionError extends ExtendableError {
   }
 }
 
-export class InvalidIDError extends ExtendableError {
+export class InvalidTransactionError extends ExtendableError {
   constructor(message: string) {
     super(message);
   }
--- a/src/coin/trx/transactionBuilder.ts
+++ b/src/coin/trx/transactionBuilder.ts
@@ -1,3 +1,4 @@
+import * as crypto from 'crypto';
 import BigNumber from 'bignumber.js';
 
 import { TransactionReceipt } from './iface';
@@ -121,8 +122,18 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     // TODO: parse the transaction raw_data_hex and compare it with the raw_data
   }
 
+  /**
+   * Validates a transaction has the right transaction id.
+   */
   validateTransaction(transaction: Transaction) {
-    // TODO: verify the transaction has the right transaction id
+    const hexBuffer = Buffer.from(transaction.toJson().raw_data_hex, 'hex');
+    const txId = crypto
+      .createHash('sha256')
+      .update(hexBuffer)
+      .digest('hex');
+    if (transaction.id != txId) {
+      throw new Error(transaction.id + ' is not a valid transaction id');
+    }
   }
 
   displayName(): string {
--- a/src/coin/trx/transactionBuilder.ts
+++ b/src/coin/trx/transactionBuilder.ts
@@ -2,7 +2,7 @@ import * as crypto from 'crypto';
 import BigNumber from 'bignumber.js';
 
 import { TransactionReceipt } from './iface';
-import { SigningError, BuildTransactionError } from '../baseCoin/errors';
+import { SigningError, BuildTransactionError, InvalidIDError } from '../baseCoin/errors';
 import { Address } from './address';
 import { BaseKey } from '../baseCoin/iface';
 import { signTransaction, isBase58Address } from './utils';
@@ -122,9 +122,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     // TODO: parse the transaction raw_data_hex and compare it with the raw_data
   }
 
-  /**
-   * Validates a transaction has the right transaction id.
-   */
+  /** @inheritDoc Specifically, checks hex underlying transaction hashes to correct transaction ID. */
   validateTransaction(transaction: Transaction) {
     const hexBuffer = Buffer.from(transaction.toJson().raw_data_hex, 'hex');
     const txId = crypto
@@ -132,7 +130,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
       .update(hexBuffer)
       .digest('hex');
     if (transaction.id != txId) {
-      throw new Error(transaction.id + ' is not a valid transaction id');
+      throw new InvalidIDError(transaction.id + ' is not a valid transaction id. Expecting: ' + txId);
     }
   }
 
--- a/src/coin/trx/transactionBuilder.ts
+++ b/src/coin/trx/transactionBuilder.ts
@@ -2,7 +2,7 @@ import * as crypto from 'crypto';
 import BigNumber from 'bignumber.js';
 
 import { TransactionReceipt } from './iface';
-import { SigningError, BuildTransactionError, InvalidIDError } from '../baseCoin/errors';
+import { SigningError, BuildTransactionError, InvalidTransactionError } from '../baseCoin/errors';
 import { Address } from './address';
 import { BaseKey } from '../baseCoin/iface';
 import { signTransaction, isBase58Address } from './utils';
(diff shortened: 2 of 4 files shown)


src/coin/trx/transactionBuilder.ts
```diff
@@ -121,8 +122,18 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+   */
   validateTransaction(transaction: Transaction) {
-    // TODO: verify the transaction has the right transaction id
+    const hexBuffer = Buffer.from(transaction.toJson().raw_data_hex, 'hex');
+    const txId = crypto
+      .createHash('sha256')
+      .update(hexBuffer)
+      .digest('hex');
+    if (transaction.id != txId) {
+      throw new Error(transaction.id + ' is not a valid transaction id');
```
argjv: You could add `txId` to the error message to show what we calculated.

test/unit/coin/trx/transactionBuilder.ts
```diff
@@ -163,5 +164,20 @@ describe('Tron', function() {
       tx.outputs[0].value.toString().should.equal('1718');
       tx.validFrom.should.equal(1571811410819);
     });
+
+    it('should catch an invalid id', () => {
+      const txJson = JSON.stringify(InvalidIDTransaction);
+      txBuilder.from(txJson);
+      txBuilder.sign({ key: FirstPrivateKey });
+      // Build calls validateTransaction()
+      should.throws(() => txBuilder.build());
```
marktoda: Can you verify that the thrown error is the expected one?

test/unit/coin/trx/transactionBuilder.ts
```diff
@@ -163,5 +164,20 @@ describe('Tron', function() {
...
+    });
+
+    it('should build a valid transaction', () => {
+      const txJson = JSON.stringify(UnsignedBuildTransaction);
+      txBuilder.from(txJson);
+      txBuilder.sign({ key: FirstPrivateKey });
+      txBuilder.build();
```
marktoda: Can you check that this did not throw or actually returned a result? It's not clear to me that mocha would consider this a failure if it threw here.

Also, is this case not already covered by another unit test above?

src/coin/trx/transactionBuilder.ts
```diff
@@ -121,8 +122,18 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     // TODO: parse the transaction raw_data_hex and compare it with the raw_data
   }
 
+  /**
+   * Validates a transaction has the right transaction id.
+   */
   validateTransaction(transaction: Transaction) {
```
marktoda: Is there any other validation you can think of to do here? I know that the original TODO only specified to check the transaction id, but there may be more that we can do here.

test/unit/coin/trx/transactionBuilder.ts
```diff
@@ -163,5 +164,20 @@ describe('Tron', function() {
...
+
+    it('should catch an invalid id', () => {
+      const txJson = JSON.stringify(InvalidIDTransaction);
+      txBuilder.from(txJson);
+      txBuilder.sign({ key: FirstPrivateKey });
+      // Build calls validateTransaction()
+      should.throws(() => txBuilder.build());
+    });
+
+    it('should build a valid transaction', () => {
```
argjv: You can remove this test since this is tested extensively in other tests like `'should build an half signed tx'` and `'should build an update account tx'`

src/coin/trx/transactionBuilder.ts
```diff
@@ -121,8 +122,18 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     // TODO: parse the transaction raw_data_hex and compare it with the raw_data
   }
 
+  /**
```
marktoda: You can just use `@inheritDoc` as the interface already defines a jsdoc

src/coin/trx/transactionBuilder.ts
marktoda: Can you use an existing or create a new custom error in `baseCoin/errors.ts` to use here? They are somewhat more specific than the generic Javascript `Error`.

src/coin/trx/transactionBuilder.ts
```diff
@@ -121,8 +122,18 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+   * Validates a transaction has the right transaction id.
```
twtaylorbitgo: Kind of pedantic, maybe could talk about the hex underlying the tx being validated? Just a thought

src/coin/trx/transactionBuilder.ts
wesgraham: Good call! Added

test/unit/coin/trx/transactionBuilder.ts
wesgraham: Great point! I see how this is redundant. Removed :)

test/unit/coin/trx/transactionBuilder.ts
wesgraham: This is the one piece I am less familiar with. Is the `InvalidIDError` definition done in the latest commit sufficient, or should I also declare that we are expecting an InvalidIDError in the `should.throws()` call here?

test/unit/coin/trx/transactionBuilder.ts
wesgraham: Resolved this by removing the test as per @argjv's suggestion!

src/coin/trx/transactionBuilder.ts
wesgraham: The Tron spec seemed to suggest to check id, signature, and expiration. I know expiration checks are in place, perhaps checking that a signature is present would also be valuable? (though I could also see this being implemented elsewhere)

src/coin/trx/transactionBuilder.ts
wesgraham: Good point! Updated

src/coin/trx/transactionBuilder.ts
wesgraham: Let me know if the additions from the recent commit are sufficient!

test/unit/coin/trx/transactionBuilder.ts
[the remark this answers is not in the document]
wesgraham: Good point - I agree. Ended up removing the case as per @argjv's suggestion!

src/coin/trx/transactionBuilder.ts
wesgraham: Noted! Added `@inheritDoc` as per suggestions above, alongside a sentence on the hex underlying the tx being validated

src/coin/trx/transactionBuilder.ts
argjv: You are correct @wesgraham 
This check can be used in unsigned transactions, so no need to check for the signatures. Also, AFAIK, the ID calculation includes the expiration time, so it is implicitly being checked here too.

src/coin/baseCoin/errors.ts
```diff
@@ -43,3 +43,9 @@ export class ExtendTransactionError extends ExtendableError {
     super(message);
   }
 }
+
+export class InvalidIDError extends ExtendableError {
```
argjv: Can you rename this to `InvalidTransactionError` ? That way it will be more generic and we can use it in other error messages.

wesgraham: Will do!

test/unit/coin/trx/transactionBuilder.ts
marktoda: Eh, it's OK -- arguably it's better to just assert that it throws, allowing us to change the error / error message later without breaking tests.

cooncesean:
@marktoda + @argjv we good to merge this?

marktoda:
@cooncesean unfortunately wes needs to sign his commits before we can merge

marktoda:
@wesgraham  -- you should be able to accomplish this simply by adding the `-S` tag to your `git commit` command. 

Here at BitGo, we require all merged commits to be signed, so most of us actually have this set by default in our git configs. You can do this with this command:
`git config --global commit.gpgsign true`

wesgraham:
Apologies on not signing! I just pushed an empty commit with a signature attached, let me know if this suffices.

wesgraham:
Apologies for the confusion! Had to force push the signature commits - let me know if everything looks alright.

wesgraham:
Turns out I had the wrong GPG key tied to my account. Everything should now be signed, hope history is not too confusing. Learned a lot about signing commits in this process haha.

PR #47, opened 2020-02-02
feat(Tron TransactionBuilder): validateKey

### LINKED ISSUE #35: Tron TransactionBuilder - Implement validateKey()
# Overview
The Tron `TransactionBuilder` class defines a method called [`validateKey()`](https://github.com/BitGo/bitgo-account-lib/blob/master/src/coin/trx/transactionBuilder.ts#L115-L118). The method currently performs zero validation on the key passed into it -- it always returns true.

Let's implement the method to perform the necessary validation.

# Definition of Done
- Implement Tron's `TransactionBuilder::validateKey()` method.
- Add test(s) to cover these changes.

dnllb:
`TransactionBuilder#validateKey` method implementation.
[The same way](https://github.com/TRON-US/tronweb/blob/master/src/index.js#L116) TronWeb verifies private key.

Closes #35


src/coin/trx/transactionBuilder.ts
```diff
@@ -113,9 +114,13 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     }
   }
 
-  validateKey(key: BaseKey) {
-    // TODO: determine valid key format
-    return true;
+  validateKey(key: BaseKey): boolean {
```
argjv: JSDocs

src/coin/trx/transactionBuilder.ts
```diff
@@ -113,9 +114,13 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+    try {
+      utils.crypto.pkToAddress(key.key);
```
argjv: This will not handle `xprv`s which are a valid format. You should replace it with `new KeyPair({ prv: key.key })` so that it test it against all our valid key types.

src/coin/trx/transactionBuilder.ts
dnlilib: Added. I thought JSDoc for a self-explainable function is not mandatory. Also added JSDoc plugin for Eslint.
After fixing linter issues across `transactionBuilder.ts` I noticed that all the `validateSomething` functions do not return any value thus the name `isKeyValid` makes more sense for `validateKey`. Just a suggestion.

src/coin/trx/transactionBuilder.ts
dnlilib: Thanks, updated 👍

src/coin/trx/transactionBuilder.ts
argjv: Thanks for the plugin, that is great!

You are absolutely right. In this case, the `return true` placeholder was misleading. This method should throw if the key is invalid somehow, not return a boolean.

It is implicitly used here in `baseTransactionBuilder.ts`:
```
  sign(key: BaseKey): void {
    // Make sure the key is valid
    this.validateKey(key);
    if (!this.transaction.canSign(key)) {
      throw new SigningError('Private key cannot sign the transaction');
    }
    this.transaction = this.signImplementation(key);
  }
```

dnllb: Updated

argjv:
This PR cannot be merged into unless each commit has been signed with a GPG key.

Here are some helpful links:
https://help.github.com/en/github/authenticating-to-github/about-commit-signature-verification
https://help.github.com/en/github/authenticating-to-github/generating-a-new-gpg-key
https://help.github.com/en/github/authenticating-to-github/signing-commits

I know it's a little painful but, it comes part and parcel with being a crypto-fintech company. No rush on it. Once you have the commits signed, I will re-accept it and merge it.

cooncesean:
Hey @dnllb -- we'd love to realize all of this great work. It's literally 95% complete -- once you sign those commits and force push, we'll go ahead and land it and discuss next steps 👍

dnllb:
@cooncesean @argjv squashed & signed

dnllb:
Conflicts resolved, [small import improvement](https://github.com/BitGo/bitgo-account-lib/pull/47/files#diff-fce46e48ba896b1b6c5b06c8e9f6577eR4)
Let's merge?

PR #50, opened 2020-02-05
Implement hash validation for Tezos

CapnMigraine:
Add validation for address, block and transaction hashes
Validate encoding, prefix and length

In Tezos, hashes are encoded using Base58Check

Ticket: BG-17820

### DIFF
--- a/src/coin/xtz/iface.ts
+++ b/src/coin/xtz/iface.ts
@@ -3,21 +3,21 @@
  */
 export type PrivateKey = {
   prv: string;
-}
+};
 
 /**
  * A Tezos public key with the sppk prefix or raw
  */
 export type PublicKey = {
   pub: string;
-}
+};
 
 /**
  * A seed to create Tezos key pairs. Must be between 16 and 64 Bytes long
  */
 export type Seed = {
   seed: Buffer;
-}
+};
 
 export type KeyPairOptions = Seed | PrivateKey | PublicKey;
 
@@ -31,4 +31,9 @@ export function isPrivateKey(source: KeyPairOptions): source is PrivateKey {
 
 export function isPublicKey(source: KeyPairOptions): source is PublicKey {
   return (source as PublicKey).pub !== undefined;
+}
+
+export interface HashType {
+  prefix: Buffer;
+  byteLength: number;
 }
--- a/src/coin/xtz/utils.ts
+++ b/src/coin/xtz/utils.ts
@@ -1,12 +1,14 @@
 import * as base58check from 'bs58check';
+import { HashType } from './iface';
 
 /**
  * Encode the payload to base58 with a specific Tezos prefix.
- * @param {Uint8Array} prefix to add to the encoded payload
+ *
+ * @param {Buffer} prefix to add to the encoded payload
  * @param {Buffer} payload to encode
  * @return {any} base58 payload with a Tezos prefix
  */
-export function base58encode(prefix: Uint8Array, payload: Buffer): string {
+export function base58encode(prefix: Buffer, payload: Buffer): string {
   const n = Buffer.alloc(prefix.length + payload.length);
   n.set(prefix);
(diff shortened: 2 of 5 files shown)


src/coin/xtz/iface.ts
```diff
@@ -32,3 +32,8 @@ export function isPrivateKey(source: KeyPairOptions): source is PrivateKey {
 export function isPublicKey(source: KeyPairOptions): source is PublicKey {
   return (source as PublicKey).pub !== undefined;
 }
+
+export interface HashType {
+  prefix: Buffer;
+  bytes: number;
```
argjv: s/bytes/byteLength/g

src/coin/xtz/iface.ts
```diff
@@ -32,3 +32,8 @@ export function isPrivateKey(source: KeyPairOptions): source is PrivateKey {
 export function isPublicKey(source: KeyPairOptions): source is PublicKey {
   return (source as PublicKey).pub !== undefined;
 }
+
+export interface HashType {
```
argjv: JSDoc or add comments next to the fields. Does `byte` include the prefix? Is this the decoded bytes length?

src/coin/xtz/utils.ts
```diff
@@ -1,82 +1,250 @@
...
-  const textPrefix = decodedText.slice(0, prefix.length);
-  // Check prefix
-  if (!textPrefix.equals(Buffer.from(prefix))) {
+/**
+ * Decode a string encoded as base58
+ *
+ * @param {String} hash - the string to decode
+ * @returns {Buffer | undefined}
+ */
+export function base58decode(hash): Buffer | undefined {
```
argjv: Missing input type

src/coin/xtz/utils.ts
```diff
@@ -1,82 +1,250 @@
...
+/**
+ * Returns whether or not the string is a valid Tezos hash of the given type
+ *
+ * @param {String} hash - the string to validate
+ * @param {HashType} hashType - the type of the provided hash
+ * @returns {Boolean}
+ */
+export function isValidHash(hash: string, hashType: HashType): boolean {
+  // Validate encoding
+  const decodedHash = base58decode(hash);
```
argjv: I'm not sure we need `base58decode`, I don't see it being referenced anywhere else. We could just do:

```
  let decodedHash;
  try {
    decodedHash = base58check.decode(hash);
  } catch (e) {
    return false;
  }
```

src/coin/xtz/iface.ts
CapnMigraine: Added comments for clarity

src/coin/xtz/iface.ts
CapnMigraine: done

src/coin/xtz/utils.ts
CapnMigraine: I removed the function, as per your suggestion in the comment below

src/coin/xtz/utils.ts
CapnMigraine: You're right, I added it as a more generic helper, but these utility functions are Tezos-specific, so I removed the function

src/coin/xtz/utils.ts
```diff
@@ -41,7 +25,12 @@ export function base58decode(hash): Buffer | undefined {
...
 export function isValidHash(hash: string, hashType: HashType): boolean {
   // Validate encoding
-  const decodedHash = base58decode(hash);
+  let decodedHash;
+  try {
+    decodedHash = base58check.decode(hash);
+  } catch (e) {
+    return false;
+  }
   if (!decodedHash) {
```
argjv: Can `base58check.decode(hash);` return undefined? If not, this `if` block is not neccesary.

CapnMigraine: it's not needed, removed

Issue #28, opened 2019-11-26
Fix TransactionBuilder instantiation in `index.ts`


marktoda:
# Overview

In index.ts, we see this instantiation:
```
  if (supportedCoins.includes(coin)) {
    return new Trx.TransactionBuilder(coins.get(coin));
  }
```
It is supposed to be able to instantiate a TransactionBuilder for whatever coin it is given, but it specifically always uses the `Trx` coin-specific transaction builder. 

Refactor this function so that it actually instantiates the transaction builder based on the coin it is passed

# Definition of Done
- Tests should still pass without change
- This function shouldn't use any `Trx`-specific code
- If we were to add another coin TransactionBuilder, and give this function the name of that coin, it should correctly instantiate it.

argjv:
I'm fixing this for Tezos

PR #52, opened 2020-02-06
BG-17815 Add Tezos resources and clean up docs

argjv:
[BG-17815](https://bitgoinc.atlassian.net/browse/BG-17815) First PR cleaning up some code documentation and adding Tezos multisig contract.

It also closes https://github.com/BitGo/bitgo-account-lib/issues/28

### DIFF
--- a/resources/xtz/multisig.ts
+++ b/resources/xtz/multisig.ts
@@ -1,0 +1,389 @@
+/**
+ * Create an origination operation for the generic multisg contract. It does not create a reveal
+ * operation for the source account.
+ *
+ * @param {string} counter Source account next counter
+ * @param {string} source Source account address
+ * @param {string} fee Fees in mutez to pay by the source account
+ * @param {string} gasLimit Maximum amount in mutez to spend in gas fees
+ * @param {string} storageLimit Maximum amount in mutez to spend in storage fees
+ * @param {string} balance New multisig account initial balance taken from the source account
+ * @param {string[]} pubKeys List of public keys of the multisig owner
+ * @return An origination operation
+ */
+export function genericMultisigOriginationOperation(
+    counter: string,
+    source: string,
+    fee: string,
+    gasLimit: string,
+    storageLimit: string,
+    balance: string,
+    pubKeys: string[]): any {
+  const walletPublicKeys: any[] = [];
+  pubKeys.forEach(pk => walletPublicKeys.push({ string: pk }));
+  return {
+    kind: 'origination',
+    counter,
+    source,
+    fee,
+    gas_limit: gasLimit,
+    storage_limit: storageLimit,
+    balance,
+    script: {
+      code: genericMultisig,
+      storage: { prim: 'Pair',
+        args: [{
+          int: '0'
+        }, {
+          prim: 'Pair',
+          args: [ {
+            int: '2' },
+            walletPublicKeys]
+        }]
+      },
+    },
+  };
+}
+
+/**
+ * Generic Multisig contract from https://github.com/murbard/smart-contracts/blob/master/multisig/michelson/generic.tz
+ */
+export const genericMultisig = [{
+  "prim": "parameter",
+  "args": [{
+    "prim": "or",
+    "args": [{ "prim": "unit", "annots": ["%default"] },
+      {
+        "prim": "pair",
+        "args": [{
+          "prim": "pair",
+          "args": [{
+            "prim": "nat",
+            "annots": ["%counter"]
+          },
+            {
+              "prim": "or",
+              "args": [{
+                "prim": "lambda",
+                "args": [{ "prim": "unit" },
+                  {
+                    "prim": "list",
+                    "args": [{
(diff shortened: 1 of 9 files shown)


resources/xtz/multisig.ts
```diff
@@ -0,0 +1,389 @@
...
+ * @return An origination operation
+ */
+export function genericMultisigOriginationOperation(
+    counter: string,
+    source: string,
+    fee: string,
+    gasLimit: string,
+    storageLimit: string,
+    balance: string,
+    pubKeys: string[]) {
```
CapnMigraine: nit: add return type

src/coin/baseCoin/baseTransaction.ts
```diff
@@ -37,13 +38,19 @@ export abstract class BaseTransaction {
    * Whether the private key can sign this transaction in its current state or not. it is possible
    * some transactions can only enforce this check after some other fields have been filled already
    * or even during build time.
-   * @param key to verify permissions on
+   *
+   * @param key Private key to verify permissions on
```
CapnMigraine: nit: use `@param {BaseKey} key Private ...`

src/index.ts
```diff
@@ -27,9 +32,9 @@ export const supportedCoins = {
  */
 export function getBuilder(coinName: string): BaseTransactionBuilder {
   const coin = coinName.toLowerCase().trim();
-  const builder = supportedCoins[coin];
-  if (builder) {
-    return builder;
+  const builderClass = coinBuilderMap[coin];
```
CapnMigraine: You're not validating `coin` as one of the `supportedCoins`

argjv: if `coinBuilderMap` does not have the ticker, then returns `undefined`, hence the `if (builderClass) { }`

resources/xtz/multisig.ts
```diff
@@ -0,0 +1,389 @@
...
+ * @return An origination operation
+ */
+export function genericMultisigOriginationOperation(
+    counter: string,
+    source: string,
+    fee: string,
+    gasLimit: string,
+    storageLimit: string,
+    balance: string,
+    pubKeys: string[]): any {
```
twtaylorbitgo: why doesn't this return a type? at least it could be bound to a poor interface

resources/xtz/multisig.ts
```diff
@@ -0,0 +1,389 @@
...
+        }]
+      },
+    },
+  };
+}
+
+/**
+ * Generic Multisig contract from https://github.com/murbard/smart-contracts/blob/master/multisig/michelson/generic.tz
+ */
+export const genericMultisig = [{
```
twtaylorbitgo: can this be just imported as a resource? it looks like valid JSON - just in case this contract type changes

src/coin/baseCoin/baseTransaction.ts
```diff
@@ -37,13 +38,19 @@ export abstract class BaseTransaction {
...
   abstract canSign(key: BaseKey): boolean;
 
   /**
-   * Returns the transaction in a valid JSON format.
+   * Return the transaction in a coin specific JSON format.
    */
   abstract toJson(): any;
+
+  /**
+   * Return the transaction in a format it can be broadcasted to the blockchain.
```
twtaylorbitgo: I feel like this could be named better - broadcastFormat or something. Package is very nebulous and makes me think of a library tbh

src/index.ts
```diff
@@ -27,9 +32,9 @@ export const supportedCoins = {
...
+  if (builderClass) {
+    return new builderClass(coins.get(coin));
```
twtaylorbitgo: from a JS perspective, it's better to check for errors first then return the normal case second. so these two statements switched around basically.

src/coin/baseCoin/baseTransaction.ts
argjv: Done

resources/xtz/multisig.ts
argjv: I'm not sure what you mean. It is in the resource folder already.

src/index.ts
argjv: Done

resources/xtz/multisig.ts
argjv: The types are in here: https://github.com/ecadlabs/taquito/blob/26f52e1dc181a50bb46c613d895a2b0fa4c85e0b/packages/taquito-rpc/src/types.ts

I just haven't decided how to import them yet since I don't want to bring extra libs just for the types. I'll fix it in the follow up Tezos PR which uses this method

twtaylorbitgo: Ok that works

resources/xtz/multisig.ts
twtaylorbitgo: Oh I see. I was thinking more like a json file. That would require some file reading or a build step. This is probably fine until you get types defining what this is

PR #54, opened 2020-02-07
Bg 17815 xtz init wallet tx builder

argjv:
[BG-17815](https://bitgoinc.atlassian.net/browse/BG-17815) Add Tezos transaction builder supporting:

* Wallet initialization transaction building
* Transaction signing
* Reveal operation support


src/coin/xtz/transaction.ts
```diff
@@ -21,25 +22,21 @@ export class Transaction extends BaseTransaction {
...
   }
 
   /**
-   * Initialize the transaction fields based on a serialized transaction.
+   * Initialize the transaction fields based on another serialized transaction.
    *
    * @param serializedTransaction Transaction in broadcast format.
    */
-  async init(serializedTransaction: string): Promise<void> {
+  async fromSerializedTransaction(serializedTransaction: string): Promise<void> {
```
CapnMigraine: the `from` makes it sound like the method should return something. 

Can we name it `initFromSerializedTransaction`?

src/coin/xtz/transaction.ts
```diff
@@ -50,41 +47,57 @@ export class Transaction extends BaseTransaction {
...
+   * Initialize the transaction fields based on another parsed transaction.
    *
-   * @param parsedTransaction A Tezos transaction object
+   * @param {ParsedTransaction} parsedTransaction A Tezos transaction object
+   * @param {string} transactionId The transaction id of the parsedTransaction if it is signed
    */
-  private recordRawDataFields(parsedTransaction: ParsedTransaction): void {
-    // TODO: make sure transaction ids cannot be calculated offline
-    this._id = '';
+  async fromParsedTransaction(parsedTransaction: ParsedTransaction, transactionId?: string): Promise<void> {
```
CapnMigraine: Same as above. Will you consider naming this one `initFromParsedTransaction`?

src/coin/xtz/transaction.ts
```diff
@@ -50,41 +47,57 @@ export class Transaction extends BaseTransaction {
...
+          await this.recordOriginationOpFields(operation as Origination, operationIndex);
+          operationIndex++;
+          break;
+        case CODEC.OP_REVEAL:
+          this.recordRevealOpFields(operation as Reveal);
           break;
         default:
-          throw new ParseTransactionError('Unsupported contract type');
+          break;
+          // throw new ParseTransactionError('Unsupported contract type');
```
CapnMigraine: leftover comment?

src/coin/xtz/transaction.ts
```diff
@@ -50,41 +47,57 @@ export class Transaction extends BaseTransaction {
...
    */
-  private recordOriginationOpFields(operation: Operation): void {
+  private async recordOriginationOpFields(operation: Origination, index: number): Promise<void> {
     this._type = TransactionType.WalletInitialization;
     this._outputs.push({
-      // TODO: use the function in eztz to calculate the originated account address and assign it to
-      // this._id (https://github.com/TezTech/eztz/blob/cfdc4fcfc891f4f4f077c3056f414476dde3610b/src/main.js#L768)
-      address: '',
+      // Kt addresses can only be calculated for signed transactions with an id
+      address: (this._id) ? await Utils.calculateOriginatedAddress(this._id, index) : '',
```
CapnMigraine: nit: replace `(this._id)` with `this._id`

src/coin/xtz/transaction.ts
```diff
@@ -95,6 +108,20 @@ export class Transaction extends BaseTransaction {
...
+   * Record the most important fields from a reveal operation.
+   *
+   * @param {Reveal} operation A reveal operation object from a Tezos transaction
+   */
+  private recordRevealOpFields(operation: Reveal): void {
+    this._type = TransactionType.AddressInitialization;
+    this._inputs.push({
+      address: operation.source,
+      // Balance + fees + max gas + max storage are paid by the source account
+      value: new BigNumber(operation.balance).plus(operation.fee)
```
CapnMigraine: nit: add dangling comma

src/coin/xtz/transaction.ts
```diff
@@ -120,6 +147,11 @@ export class Transaction extends BaseTransaction {
 
     const signedTransaction = await signer.sign(encodedTransaction, new Uint8Array([3]));
     this._encodedTransaction = signedTransaction.sbytes;
+
+    // The transaction id can only be calculated for signed transactions
+    this._id = await Utils.calculateTransactionId(this._encodedTransaction);
+    await this.fromParsedTransaction(this._parsedTransaction, this._id);
```
CapnMigraine: isn't the transaction object already initialized here?

src/coin/xtz/transactionBuilder.ts
```diff
@@ -0,0 +1,293 @@
...
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<Transaction> {
+    // If the from() method was called, use the serialized transaction as a base
+    if (this._serializedTransaction) {
+      await this.transaction.fromSerializedTransaction(this._serializedTransaction);
+      // TODO: make changes to the transaction if any extra parameter has been set then sign it
+      return this.transaction;
+    }
+
+    let contents: Operation[] = [];
```
CapnMigraine: use `const`

src/coin/xtz/transactionBuilder.ts
```diff
@@ -0,0 +1,293 @@
...
+        const revealOp = revealOperation(
+          this._counter.toString(),
+          this._sourceAddress,
+          DEFAULT_FEE.REVEAL.toString(),
+          DEFAULT_GAS_LIMIT.REVEAL.toString(),
+          DEFAULT_STORAGE_LIMIT.REVEAL.toString(),
+          this._amount.toString(),
+          this._sourceKeyPair!.getKeys().pub);
+          contents.push(revealOp);
+          this._counter = this._counter.plus(1);
```
CapnMigraine: nit: remove extra indentation

src/coin/xtz/transactionBuilder.ts
```diff
@@ -0,0 +1,293 @@
...
+        contents.push(originationOp);
+        break;
+      case TransactionType.Send:
+        throw new BuildTransactionError('Send is not supported yet');
+      default:
+        throw new BuildTransactionError('Unsupported transaction type');
+    }
+    const parsedTransaction = {
+      branch: this._blockHeader,
+      contents
```
CapnMigraine: nit: add dangling comma

src/coin/xtz/transactionBuilder.ts
```diff
@@ -0,0 +1,293 @@
...
+
+    return Promise.resolve(this.transaction);
+  }
+
+  /**
+   * The type of transaction being built.
+   *
+   * @param {TransactionType} type
+   */
+  type(type: TransactionType) {
```
CapnMigraine: nit add return comment for this one and all the other methods below

src/coin/xtz/transaction.ts
```diff
@@ -50,41 +47,57 @@ export class Transaction extends BaseTransaction {
...
+      this._id = '';
+    }
     this._parsedTransaction = parsedTransaction;
-    parsedTransaction.contents.forEach(operation => {
+    let operationIndex = 0;
+    parsedTransaction.contents.forEach(async operation => {
       switch (operation.kind) {
         case CODEC.OP_ORIGINATION:
-          this.recordOriginationOpFields(operation);
+          await this.recordOriginationOpFields(operation as Origination, operationIndex);
```
CapnMigraine: Shouldn't `operationIndex` hold the index of the origination operation within the array of operations?

I think the counter should increment regardless of the type of operation.

If my previous statement is correct, then you can use the counter of the iteration as index.

e.g.
```
...
    this._parsedTransaction = parsedTransaction;
    parsedTransaction.contents.forEach(async (operation, index) => {
      switch (operation.kind) {
        case CODEC.OP_ORIGINATION:
          await this.recordOriginationOpFields(operation as Origination, index);
          break;
...

File: src/coin/xtz/keyPair.ts [hunk 12]
```diff
@@ -116,15 +116,14 @@ export class KeyPair {
   }
 
   /**
-   * Get a public address in the specified format, or in base58 if none is provided.
+   * Get a public address.
    */
   getAddress(): string {
-    const { pub } = this.getKeys();
+    const pub = this.keyPair.getPublicKeyBuffer();
```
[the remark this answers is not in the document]
CapnMigraine: nit: Use the same method to get the pub key buffer

`this.keyPair.getPublicKeyBuffer();` returns the same value as `this.keyPair.Q.getEncoded(true);`, used in `getKeys()`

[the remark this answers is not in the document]
argjv: Yes. This was the bug. `this.getKeys().pub` is in Tezos pubkey format when it should have been in its raw generic one.

src/coin/xtz/transaction.ts
argjv: Some of the fields yes, but we are passing the calculated id, which will update fields the originated addresses fields.

src/coin/xtz/transaction.ts
argjv: I'm not entirely sure of this actually. I know that the index is used to produce different addresses for multiple origination operations in the same transactions since the address is a hash of the transaction + index.
So far, this produced the right address with index zero, even when it is the second operation in the transaction.

src/coin/baseCoin/baseTransactionBuilder.ts
```diff
@@ -62,9 +62,9 @@ export abstract class BaseTransactionBuilder {
    *
    * @return valid coin specific transaction (signed or unsigned)
    */
-  build(): BaseTransaction {
+  async build(): Promise<BaseTransaction> {
```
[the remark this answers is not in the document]
argjv: Good point. This is a breaking change for sure, however, I kinda assumed all operations that called this method were coroutines.
I'm not sure I can make this method `sync` since we rely on sodium lib which has some async methods.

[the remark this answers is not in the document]
CapnMigraine: I checked the platform logic, it shouldn't be a problem since the `signTxRemoteV2` method is also a coroutine.

src/coin/xtz/transaction.ts
CapnMigraine: do you think it's possible to have an operation group where the the origination is not in index 0? 

If that's the case it would be nice to have a test for it

src/coin/xtz/transaction.ts
CapnMigraine: remove if it's not needed

src/coin/xtz/iface.ts
```diff
@@ -40,15 +41,31 @@ export interface HashType {
...
   kind: string;
+  counter: string;
   balance: string;
   source: string;
   fee: string;
   gas_limit: string;
   storage_limit: string;
+}
+
+export interface Reveal extends Operation{
```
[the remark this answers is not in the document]
cooncesean: Can we automate this type of feedback with a linter + pre-commit hook?

src/coin/xtz/transaction.ts
argjv: Yes, when we have a reveal operation, the origination is done after that. There is a test for this already (reveal + origination ops in the same TX)

src/coin/xtz/iface.ts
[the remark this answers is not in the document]
CapnMigraine: It's currently broken, here's a fix for it: https://github.com/BitGo/bitgo-account-lib/pull/59

PR #59, opened 2020-02-12
Fix lint pre-commit hook and scripts

CapnMigraine:
Run lint-staged pre-commit hook
Fix lint and lint-fix scripts

Ticket: BG-18037


src/coin/trx/iface.ts
```diff
@@ -71,7 +71,7 @@ export interface TransactionReceipt {
 export interface RawData {
   expiration: number;
   timestamp: number;
-  contractType: ContractType;
+  contractType?: ContractType;
```
CapnMigraine: @argjv - not 100% sure on this one. `UnsignedBuildTransaction` in `test/resources/trx:13` doesn't have a `contractType`

src/coin/xtz/transaction.ts
```diff
@@ -80,7 +80,7 @@ export class Transaction extends BaseTransaction {
           break;
         default:
           break;
-          // throw new ParseTransactionError('Unsupported contract type');
+        // throw new ParseTransactionError('Unsupported contract type');
```
argjv: My bad. Can you remove this line?

src/coin/trx/iface.ts
argjv: I think it's fine. It is missing sometimes

src/coin/xtz/transaction.ts
CapnMigraine: done

PR #48, opened 2020-02-04
Tron TransactionBuilder: Improve validation during transaction parsing

JackyW47:
## Description

This pull request is to address the issue proposed in #34 . Changes are as follows:

* Added validation logic in `TransactionBuilder.fromImplementation`
* Fixed related test case - Now if a `rawTransaction` is invalid, error will be thrown during `txBuilder.from`

## Things to note

Currently the unit test is not complete. It has not covered the case that raw_data does not equal to raw_data_hex since prerequisite #46 is still under review. Will implement another negative test after #46 is merged to master.

PR #46, opened 2020-01-31
Validate trx

leviathan-mist:
Closes #36 

Adding methodology for validation check for a raw transaction

PR #61, opened 2020-02-26
BG-18444 Fix multisig send

argjv:
Add contract counter parameter to the generated data to sign (`genericMultisigDataToSign `)

### DIFF
--- a/src/coin/trx/transactionBuilder.ts
+++ b/src/coin/trx/transactionBuilder.ts
@@ -38,7 +38,6 @@ export class TransactionBuilder extends BaseTransactionBuilder {
    * @returns {Transaction} Tron transaction
    */
   protected fromImplementation(rawTransaction: TransactionReceipt | string): Transaction {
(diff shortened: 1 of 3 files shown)


src/coin/trx/transactionBuilder.ts
```diff
@@ -38,7 +38,6 @@ export class TransactionBuilder extends BaseTransactionBuilder {
    * @returns {Transaction} Tron transaction
    */
   protected fromImplementation(rawTransaction: TransactionReceipt | string): Transaction {
-    // TODO: add checks to ensure the raw_data, raw_data_hex, and txID are from the same transaction
```
CapnMigraine: don't we need those anymore?

argjv: No. We fixed this in https://github.com/BitGo/bitgo-account-lib/pull/46 

More details here: https://github.com/BitGo/bitgo-account-lib/pull/48

PR #62, opened 2020-02-26
Fix compatibility issues with Babel

CapnMigraine:
Change module imports
Move class member initialization to controllers
Cleanup by eslint

Ticket: BG-18455

### DIFF
--- a/src/coin/trx/utils.ts
+++ b/src/coin/trx/utils.ts
@@ -1,6 +1,5 @@
-import * as assert from 'assert';
+import assert from 'assert';
 import * as tronweb from 'tronweb';
-import { HDNode, ECPair } from 'bitgo-utxo-lib';
 import { protocol } from '../../../resources/trx/protobuf/tron';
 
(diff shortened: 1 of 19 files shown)


src/coin/trx/utils.ts
```diff
@@ -100,11 +142,11 @@ export function decodeTransaction(hexString: string): RawData {
   switch (rawTransaction.contracts[0].parameter.type_url) {
     case 'type.googleapis.com/protocol.TransferContract':
       contractType = ContractType.Transfer;
-      contract = this.decodeTransferContract(rawTransaction.contracts[0].parameter.value);
+      contract = exports.decodeTransferContract(rawTransaction.contracts[0].parameter.value);
```
argjv: Is this the Typescript way of doing it?

CapnMigraine: Babel blows up with `this`. This way is also valid. It's covered by the Tron TransactionBuilder tests

PR #71, opened 2020-03-17
Fix signature override

argjv:
When signing a Tezos half signed transaction, the account lib generates a new array of signatures to replaces the existing one. Old signatures are discarded while they should be merged if any exist.

TICKET: BG-19070

### DIFF
--- a/src/coin/xtz/transactionBuilder.ts
+++ b/src/coin/xtz/transactionBuilder.ts
@@ -375,7 +375,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     const signatures: { signature: string; index: number }[] = [];
     // Generate the multisig contract signatures
     for (let i = 0; i < this._multisigSignerKeyPairs.length; i++) {
-      const signature = await Utils.sign(this._multisigSignerKeyPairs[i].key, packedData, new Uint8Array());
+      const signature = await Utils.sign(this._multisigSignerKeyPairs[i].key, packedData, new Uint8Array(0));
       const index = this._multisigSignerKeyPairs[i].index || i;
       signatures.push({ signature: signature.sig, index });
     }
--- a/src/coin/xtz/multisigUtils.ts
+++ b/src/coin/xtz/multisigUtils.ts
@@ -178,7 +178,7 @@ function genericMultisigTransferParams(
  * @param {IndexedSignature[]} signatures List of transactions and their order
  */
 export function updateMultisigTransferSignatures(transaction: TransactionOp, signatures: IndexedSignature[]) {
-  transaction.parameters.value.args[1] = buildSignatures(signatures);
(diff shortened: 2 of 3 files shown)


resources/xtz/multisig.ts
```diff
@@ -178,21 +178,27 @@ function genericMultisigTransferParams(
...
  * @returns {any[]} List of signatures in the right order
  */
-function buildSignatures(signatures: IndexedSignature[], m: number = DEFAULT_M) {
-  const transactionSignatures: any[] = [];
-  // Initialize the array with empty signatures
-  for (let i = 0; i < m; i++) {
-    transactionSignatures.push({ prim: 'None' });
+function buildSignatures(signatures: IndexedSignature[], existingSignatures = [], m: number = DEFAULT_M) {
+  // Initialize the array with the existing signatures and/or empty objects
+  const transactionSignatures: any[] = existingSignatures;
```
CapnMigraine: nit: can we use `object` as `existingSignature`'s type instead of `any`?

resources/xtz/multisig.ts
```diff
@@ -178,21 +178,27 @@ function genericMultisigTransferParams(
...
+  const size = existingSignatures.length;
+  if (size <= m) {
+    for (let i = size; i < m; i++) {
+      transactionSignatures.push({ prim: 'None' });
+    }
+  } else {
+    throw new Error('Too many signatures. Expected less than ' + m + ' got ' + size);
   }
```
CapnMigraine: readability-nit: convert check to guard

```suggestion
  if (size > m) {
    throw new Error('Too many signatures. Expected less than ' + m + ' got ' + size);
  }
  for (let i = size; i < m; i++) {
    transactionSignatures.push({ prim: 'None' });
  }
```

src/coin/xtz/transactionBuilder.ts
```diff
@@ -375,7 +375,7 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     const signatures: { signature: string; index: number }[] = [];
     // Generate the multisig contract signatures
     for (let i = 0; i < this._multisigSignerKeyPairs.length; i++) {
-      const signature = await Utils.sign(this._multisigSignerKeyPairs[i].key, packedData, new Uint8Array());
+      const signature = await Utils.sign(this._multisigSignerKeyPairs[i].key, packedData, new Uint8Array(0));
```
CapnMigraine: is there a difference between `new Uint8Array()` and `new Uint8Array(0)`?

argjv: Not really, but `tsc` was complaining about it

resources/xtz/multisig.ts
argjv: If I do that it gets replaced by `Record<string, any>[]`. Is that correct?

PR #79, opened 2020-03-21
Remove coin specific scripts from package.json

aiolos404:
I proposal a way to remove coin specific scripts from package.json

1. Creat **coins.config.json** file for coins configurations. Each coin has coins.${_short_name_}.packages as key to retrieve its essential package scripts.
2. New script `build-coin` is added in package.json. [jq](https://github.com/stedolan/jq) is used in the script so it should be installed in system.
3. `COIN=TRX npm run build-coin` is an example command to install the coin specified packages and other essential packages. `TRX` will be passed into script and the packages information in **coins.config.json** will be read into **package.json**.
4. Build process will start when every information is ready.
5. When a new type of coin is ready to be added, just simply add its specific packages information in **coins.config.json** and then call with `COIN=_COINNANE_ npm run build-coin` command.

Hope this clarify my thought and code. Please let me know if there's any question.

### DIFF
--- a/coins.config.json
+++ b/coins.config.json
@@ -1,7 +1,7 @@
 {
   "coins": {
     "TRX":{
-      "packages":"npm run gen-protobuf && npm run gen-protobufts"
+      "packages":"node node_modules/protobufjs/cli/bin/pbjs -t static-module -w commonjs -o ./resources/trx/protobuf/tron.js ./resources/trx/protobuf/Discover.proto ./resources/trx/protobuf/Contract.proto ./resources/trx/protobuf/tron.proto && node node_modules/protobufjs/cli/bin/pbts -o ./resources/trx/protobuf/tron.d.ts ./resources/trx/protobuf/tron.js"
     }
   }
 }
--- a/coins.config.json
+++ b/coins.config.json
@@ -1,0 +1,7 @@
+{
+  "coins": {
+    "TRX":{//Tron coin
+      "packages":"npm run gen-protobuf && npm run gen-protobufts"
+    }
+  }
+}
--- a/coins.config.json
+++ b/coins.config.json
@@ -1,6 +1,6 @@
 {
   "coins": {
-    "TRX":{//Tron coin
+    "TRX":{
       "packages":"npm run gen-protobuf && npm run gen-protobufts"
     }
   }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "webpack-test": "webpack --progress --colors --env.test",
     "webpack-prod": "webpack --progress --colors --env.prod",
     "build": "tsc && cp -r ./resources ./dist",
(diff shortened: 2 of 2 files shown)


CapnMigraine:
Hey @aiolos404, thanks for putting up. 
I have a couple of notes: 

We want to clean up the scripts section from coin-specific tasks like `gen-protobuf` and `gen-protobufts` too. Those could also be moved to a separate file.

We want to be able to scale this solution for other coin-specific tasks in the future. The `coins.config.json` config helps to that end. Have you thought about a config or script per coin implementation?

Currently, the scripts required by Tron run on `npm install`,  we want to keep that behavior instead of adding more steps to the setup. With this change, `build-coin` would have to run before `build`.

Can you integrate the `build-coin` command so it doesn't require a coin for argument? 

Do you see a way around the new npm dependency? e.g. a bash script that runs sub-scripts defined in external per-coin files.

aiolos404:
Hi @CapnMigraine, thanks for the note. Please check out my thoughts based on your note as following:
> We want to clean up the scripts section from coin-specific tasks like `gen-protobuf` and `gen-protobufts` too. Those could also be moved to a separate file.

It make sense to me. All the assistant tasks can be integrated into a separate file. I will think about that.






> We want to be able to scale this solution for other coin-specific tasks in the future. The `coins.config.json` config helps to that end. Have you thought about a config or script per coin implementation?

My thought was to keep single config for all coins if we do not have many coins to support. But in the long run, if we supports many coins, then its better to set multiple configs for each coin in same folder, like `configs/coins/TRX.config.json` or `configs/coins/XTZ.config.json` if there are many coins supported and their configs cannot be integrated into single config file due to the size and LOC. However, in both way, the script can be invoked through parameters in package.json.



> Currently, the scripts required by Tron run on `npm install`, we want to keep that behavior instead of adding more steps to the setup. With this change, `build-coin` would have to run before `build`.

I think this can be done through update a script like `  "scripts": {
    "foo": "npm run bar"
  }` to guarantee the execution order of script.



> Can you integrate the `build-coin` command so it doesn't require a coin for argument?

In this case, how can we specify which coin setting user would set with the environment? I was planning to set command like `build:TRX` and `build:XTZ` for specific coins. However, if we are planning to support many more coins in future, we may not want to integrate all these separate commands for each coins. 
The solution I propose here are: 1, use argument to define the coin setting, 2, write another separate file to store those pre-set setting commands in another config/script, which can be invoked through `npm run` in script of package.json.



> Do you see a way around the new npm dependency? e.g. a bash script that runs sub-scripts defined in external per-coin files.

To be honest, I read through the closes issues and found there's already a pull request use bash script to finish this issue. I think that's a good idea but need to set bash script for each coin. My  design idea is to reduce LOC and increase the code reuse as much as possible. That's the basis I made this design.


Thank you again for the feedback. I think my idea may not be a best solution due to lack of discussion before I start. However, since this is a code challenge, I think I can create solution in a quickly iteratively way with more communication with you all.

aiolos404:
Since we agreed #81 is better solution, this pull request is closed.

Issue #27, opened 2019-11-26
Remove Coin-Specific Scripts from Package.json (1)


marktoda:
# Overview

There are several TRX-specific commands in the `package.json` scripts for this library. For example, `build` calls `gen-protobuf` which really only relates to TRX. To avoid adding extra `&& xyz` steps for each new coin that has a special build step, let's refactor the coin-specific scripts out of `package.json`.  

This should be done in such a way that if any new coin has a special step, we can add it without having to update `package.json`.

manikey123:
Based on research the approach is below. Let me know feedback
create a bash file (newcoinscripts.sh) move the scripts in it.
Add the script in package.json
 "build":"./bin/newcoinscripts.sh"

marktoda:
That sounds great! Thanks

manikey123:
Hi
below are the changes. Let me know your feedback
https://github.com/BitGo/bitgo-account-lib/pull/58/files

PR #81, opened 2020-03-25
Add coin config script for TRX coin

aiolos404:
This pull request is an alternative way to fix issue https://github.com/BitGo/bitgo-account-lib/issues/27, based on the discussion about "a config or script per coin implementation" in https://github.com/BitGo/bitgo-account-lib/pull/79#issuecomment-603894410

Each coin will have separate config file in `configs/` and can be invoked by passing `COIN` argument in `npm install`, e.g. `COIN=TRX npm install`


package.json
```diff
@@ -8,16 +8,15 @@
     "webpack-dev": "webpack --progress --colors --env.dev",
     "webpack-test": "webpack --progress --colors --env.test",
     "webpack-prod": "webpack --progress --colors --env.prod",
-    "build": "npm run gen-protobuf && npm run gen-protobufts && tsc && cp -r ./resources ./dist",
+    "build": "tsc && cp -r ./resources ./dist",
+    "postinstall": "cmd=$(cat ./configs/$COIN.config.json | jq -r .coins.packages) | $cmd && tsc && cp -r ./resources ./dist",
```
CapnMigraine, line 12: the `build` script runs before `postinstall`. The commands in `build` should be executed after the TRX script puts the files in `./resources`

aiolos404: This can be solved by update/remove the prepare script. commit d33fbc4 removed postinstall to separate the install and build processes. User can first run `npm install `to set up dependencies and then run `npm run build:All` to go through and execute all coin configures, or just run script for a specific coin, for example: `npm run build:TRX`.

CapnMigraine: The two goals we want to achieve with this change are:
* to be able to move coin-specific setup steps from `package.json`, and 
* to keep the installation process simple (i.e it should not require any extra commands).

All of the dependencies should be installed on `npm install`

package.json
```diff
@@ -8,20 +8,23 @@
...
     "webpack-test": "webpack --progress --colors --env.test",
     "webpack-prod": "webpack --progress --colors --env.prod",
-    "build": "npm run gen-protobuf && npm run gen-protobufts && tsc && cp -r ./resources ./dist",
+    "build": "tsc && cp -r ./resources ./dist",
+    "prepare:TRX" : "cmd=$(cat ./configs/TRX.config.json | jq -r .coins.packages) | $cmd",
+    "prepare:XTZ" : "cmd=$(cat ./configs/XTZ.config.json | jq -r .coins.packages) | $cmd",
+    "prepare:All": "npm run prepare:TRX && npm run prepare:XTZ ",
+    "build:TRX": "npm run prepare:TRX && npm run build",
+    "build:XTZ": "npm run prepare:XTZ && npm run build",
+    "build:All": "npm run prepare:TRX && npm run prepare:XTZ && npm run build",
```
CapnMigraine, line 17: We want the scripts section as clean as possible. There should be only one script that takes care of all coin-specific dependencies

package.json
aiolos404: I removed all coin specific json configurations and add a coins.config.sh to keep all the coin specific scripts. In this way, all coin-specific setup steps are removed from `package.json` and only `npm install` is needed in installation. Please refer to commit 1e680ef

package.json
aiolos404: In commit 1e680ef, all coin-specific scripts are removed from `package.json`. If a new coin is added into product, its setting can be simply added in coins.config.sh in order to make sure its dependencies can be installed in the install/build process.

aiolos404:
Been waiting for the review too long. Thank you for the reviewing process.

Issue #80, opened 2020-03-24
Add get signatures method


argjv:
Transfer signatures are buried deep into the transfer script. We need a method to get them since users may want to verify the signatures.

For instance, to get the 1st signature, user must get the JSON object and search for: `contents[i].parameters.value.args[1][0].args[0].string`

# Solution
Write a method to return multisig transfer signatures as `IndexedSignature[]`.

PR #83, opened 2020-03-26
Add methods to work with nested signatures

argjv:
In this PR:
* Add a method to get the signatures and indexes from a multisig transfers (Solves https://github.com/BitGo/bitgo-account-lib/issues/80).
* Add a method to get the indexes of different operations. This is used to inspect the transaction and properly sign.

### DIFF
--- a/src/coin/xtz/transaction.ts
+++ b/src/coin/xtz/transaction.ts
@@ -20,6 +20,7 @@ import * as Utils from './utils';
 export class Transaction extends BaseTransaction {
   private _parsedTransaction?: ParsedTransaction; // transaction in JSON format
   private _encodedTransaction?: string; // transaction in hex format
+  private _source: string;
 
   /**
    * Public constructor.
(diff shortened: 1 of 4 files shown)


src/coin/xtz/transaction.ts
```diff
@@ -217,4 +221,36 @@ export class Transaction extends BaseTransaction {
...
   }
+
+  /**
+   * Get the signatures for the given multisig transfer,
+   *
+   * @param {number} transferIndex The transfer script index in the Tezos transaction
+   * @returns {IndexedSignature[]} A list of signatures with their index inside the multisig transfer
+   *      script
+   */
+  getTransferSignature(transferIndex = 0): IndexedSignature[] {
```
CapnMigraine: nit: since this function returns a list of signatures, can we name it `getTransferSignatures`?

src/coin/xtz/transaction.ts
```diff
@@ -217,4 +221,36 @@ export class Transaction extends BaseTransaction {
...
+    return getMultisigTransferSignatures(this._parsedTransaction.contents[transferIndex] as TransactionOp);
+  }
+
+  /**
+   * Get the list of index per tezos transaction type. This is useful to locate specific operations
+   * within the transaction and verify or sign them.
+   *
+   * @returns {{[p: string]: number[]}} List of indexes where the key is the transaction kind
+   */
+  getIndexByTransactionType(): { [kind: string]: number[] } {
```
CapnMigraine: nit: same here, I think `getIndexesByTransactionType` is more accurate

PR #86, opened 2020-04-14
Added support for eth, added mapping from rsk to eth.

gonzaalvarez70:
Added initial files for eth support and a mapping in the index file to
use eth when the coin type is 'rsk'.
Added support for WalletInitialization transaction in eth.
Added signature method avoiding ethereumtx-js library.
Added web3 library for some features, will be removed later.
Added unit tests.


src/coin/eth/address.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseAddress } from '../baseCoin/iface';
+
+/**
+ * Base 58 address with tz2 prefix (secp256k1)
```
argjv, line 4: Update doc. This one is Tezos specific.

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  /**
+   * This will contain the data of the message or the init of a contract
+   */
+  data?: BufferLike;
+
+  /**
+   * EC recovery ID.
+   */
+  v?: BufferLike;
```
argjv, line 146: You can extend `SigData` to dedup these params

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,186 @@
+import BN = require('bn.js');
```
argjv, line 1: Let's use `https://www.npmjs.com/package/bignumber.js` instead since we use it internally already. The least dependencies we bring in, the better.

package.json
```diff
@@ -42,7 +42,12 @@
     "elliptic": "^6.5.2",
     "libsodium-wrappers": "^0.7.6",
     "protobufjs": "^6.8.9",
-    "tronweb": "^2.7.2"
+    "tronweb": "^2.7.2",
+    "web3": "^1.2.6",
+    "keccak": "^1.0.2",
+    "rlp": "^2.0.0",
+    "keccak256": "^1.0.0",
```
argjv, line 49: Why do we need this dependency? Isn't `keccak` enough?

package.json
```diff
@@ -42,7 +42,12 @@
...
+    "underscore": "^1.9.1"
```
argjv, line 50: Seems like we use this only for the `isNumber()` and `isString()` method. We can do those checks without an external lib.

src/coin/eth/address.ts
gonzaalvarez70: Done, removed the unused file

src/coin/eth/iface.ts
gonzaalvarez70: Nice catch, done.

src/coin/eth/iface.ts
gonzaalvarez70: I tried but we are using the function toArrayLike and it's not something trivial to implement, we could add a new story to solve this issue.

package.json
gonzaalvarez70: We are using both right now, but I think that we could change it to just use keccak256, but it will need some time, maybe we can create another story to work in parallel.

package.json
gonzaalvarez70: Sure, I changed it, please check if the new implementation seems correct. It's on line #109 from utils.ts

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,186 @@
...
+
+export interface Account {
+  publicKey: string;
+  privateKey: string;
+  address: {
+    hex: string;
+  };
+}
+
+export interface ParsedTransaction {
```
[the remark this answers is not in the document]
gonzaalvarez70: Done, thanks

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,186 @@
...
+  branch: string;
+  contents: Operation[];
+}
+
+export interface Operation {
```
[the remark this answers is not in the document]
gonzaalvarez70: Done, thanks

src/coin/eth/signature.ts
```diff
@@ -0,0 +1,316 @@
...
+ * @param privateKey
+ */
+export async function signTx(data: TxData = {}, chainId: number, privateKey: Buffer): Promise<string> {
+  const tx = new Transaction(data, chainId);
+  tx.sign(privateKey);
+  const serializedTx = tx.serialize();
+  return serializedTx.toString('hex');
+}
+
+export class Transaction {
```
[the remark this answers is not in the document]
gonzaalvarez70: It's based on ethereumjs-tx

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,315 @@
...
+}
+
+/**
+ * Format transaction to be signed
+ *
+ * @param transactionData
+ */
+function formatTransaction(transactionData: TxData): TxData {
+  return {
+    gasLimit: web3.utils.toHex(transactionData.gasLimit as number),
```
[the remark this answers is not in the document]
gonzaalvarez70: It doesn't work right away, but we will investigate how to replace this. The issue comes when the numbers are converted to hex.

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,315 @@
...
+import { signTx } from './signature';
+import { KeyPair } from './keyPair';
+import { walletSimpleByteCode, walletSimpleAbi, rskDomain } from './walletUtil';
+import keccak256 from 'keccak256';
+import BN from 'bn.js';
+import Web3 from 'web3'; //TODO: Remove this dependency when it's unused
+import _ from 'underscore';
+
+export type ByteArray = number[];
+const web3 = new Web3(rskDomain);
```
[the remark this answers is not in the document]
gonzaalvarez70: We can use the empty constructor, I just figured out that, but in order to use the methods we must initialize the object using new.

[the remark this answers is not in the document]
argjv: Empty constructor should be fine as long as it guarantees it does not connect to anything.

src/coin/eth/enum.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Contract type describes a class of contract in a Transaction. This structure is used to identify what the Contract is.
+ */
+export enum ContractType {
+  /**
+   * This is a transfer of TRX contract.
```
marktoda, line 6: `TRX`?

src/coin/eth/enum.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Contract type describes a class of contract in a Transaction. This structure is used to identify what the Contract is.
```
marktoda, line 2: "Contract Type" is a little confusing for me and likely others. Maybe "TransactionType" makes more sense, as `send` and `wallet initialization` are both different types of transactions that a BitGo user may send, but they both use the same contract (the multisig wallet contract). There are also `Forwarder creation`, `token send`, `token flush` transaction types as well (though no need to preemptively add those, we will when needed)

marktoda: To be clear, this refers to the `enum` name on line 4, not just the comment

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,172 @@
+import BN = require('bn.js');
+import { BaseFee } from '../baseCoin/iface';
+
+/**
+ * An Ethereum private key in extended or raw format
+ */
+export type PrivateKey = {
```
marktoda, line 7: A lot of the stuff in this file seems copy/pasted from `src/coin/trx/iface.ts` -- any way we can dedupe these?

marktoda: keeping in mind that stuff like key management, seed management, etc. will be the same for many coins but not all

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,172 @@
...
+  /**
+   * EC signature parameter.
+   */
+  s?: BufferLike;
+}
+
+/**
+ * A transaction's data.
+ */
+export interface TxData extends SigData {
```
marktoda, line 113: This is the same type as defined in `ethereumjs-tx` if i recall correctly -- why don't we just pull in that dependency? We already use it in many places across BitGo repos

src/coin/eth/keyPair.ts
```diff
@@ -0,0 +1,123 @@
...
+      throw new Error('Unsupported private key');
+    }
+  }
+
+  /**
+   * Build a Hierarchical Deterministic node or an ECPair from a public key.
+   *
+   * @param {string} pub - An extended, compressed, or uncompressed public key
+   */
+  private recordKeysFromPublicKey(pub: string): void {
```
marktoda, line 61: These functions look the exactly same as those in `trx` keyPair.ts as well.. Can we refactor to dedupe them?

src/coin/eth/keyPair.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+    return result;
+  }
+
+  /**
+   * Get a public address
+   *
+   * @returns {string} The address derived from the private key or public key
+   */
+  getAddress(): string {
```
marktoda, line 116: `ethUtil.pubToAddress` will do most of this lift for you. `ethereumjs-util` is the package name

src/coin/eth/signature.ts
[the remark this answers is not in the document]
marktoda: Why don't we just pull in `ethereumjs-tx` as a dependency instead of copying in this whole file and types I mentioned above? Helps keep the code down and makes it easier to pull in updates / bugfixes through `npm`. We already use `ethereumjs-tx` and `ethereumjs-util` in several repos

src/coin/eth/transaction.ts
```diff
@@ -0,0 +1,80 @@
+/**
+ * Ethereum transaction model
+ */
+import { BaseTransaction, TransactionType } from '../baseCoin';
+import { BaseKey } from '../baseCoin/iface';
+import { InvalidTransactionError } from '../baseCoin/errors';
+import { KeyPair } from './keyPair';
+import { TxData } from './iface';
+import { Utils } from './';
+import { BaseCoin as CoinConfig } from '@bitgo/statics/dist/src/base';
```
marktoda, line 10: I think you can just import from `@bitgo/statics`, no need to path into `/dist`

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+    super(_coinConfig);
+    this._type = TransactionType.Send;
+    this._counter = 0;
+    this._walletOwnerAddresses = [];
+    this.transaction = new Transaction(this._coinConfig);
+  }
+
+  /** @inheritdoc */
+  protected async buildImplementation(): Promise<BaseTransaction> {
+    if (this._type != TransactionType.WalletInitialization) {
```
marktoda, line 45: Is this check needed? You could have a `default` in the switch below that throws this error

src/coin/eth/utils.ts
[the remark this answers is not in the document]
marktoda: `ethereumJs-util` [bufferToHex](https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/modules/_bytes_.md#const-buffertohex)? There is definitely a better way to convert to hex string

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * Returns the smart contract encoded data
+ *
+ * @param {string[]} addresses - the contract signers
+ * @returns {string} - the smart contract encoded data
+ */
+export function getContractData(addresses: string[]): string {
+  const contract = new web3.eth.Contract(walletSimpleAbi);
```
marktoda, line 51: Since you already store the `walletSimplebyteCode` there's no need to use `web3.eth.contract` here. If I understand correctly, you're trying to get the `bytecode` that gets used with the deploy transaction, correct? That should be just the `walletSimpleByteCode`, no need for this.

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * Checks if the given string is a checksummed address
+ *
+ * @function checkAddressChecksum
+ * @param {string} address the given HEX address
+ * @returns {boolean}
+ */
+function checkAddressChecksum(address: string): boolean {
```
marktoda, line 85: `ethereumjs-util` has a function for this

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,314 @@
...
+export type ByteArray = number[];
+const web3 = new Web3(rskDomain);
+
+/**
+ * Signs the transaction using the Eth elliptic curve
+ *
+ * @param transactionData
+ * @param keyPair
+ */
+export async function sign(transactionData: TxData, keyPair: KeyPair): Promise<string> {
```
marktoda, line 21: This only works for signing wallet creation txs (which `BitGo-account-lib` usually won't be signing, since they get signed in the HSM). When signing `sendMultisig` transactions, the key signs an operation hash, see https://github.com/BitGo/eth-multisig-v2/blob/df532e9010a0dc1ad125130da408152c12986f87/contracts/WalletSimple.sol#L130"

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,314 @@
...
+ * @param str
+ */
+export function isHexPrefixed(str: string): boolean {
+  return str.slice(0, 2) === '0x';
+}
+
+/**
+ * @param str
+ */
+export function stripHexPrefix(str: string): string {
```
marktoda, line 199: This is also in `ethUtil`

marktoda: and lots of the methods below..

src/coin/eth/enum.ts
gonzaalvarez70: Changed, thanks!

src/coin/eth/signature.ts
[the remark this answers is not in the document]
gonzaalvarez70: Sounds good, we were just trying to avoid using `ethereumjs-tx` to perform the signature without adding an external library.

src/coin/eth/enum.ts
gonzaalvarez70: Yes, we are actually using the base enum class, so we can delete this file. Thanks for the comment.

src/coin/eth/iface.ts
gonzaalvarez70: Sure, those were copied in both `trx` and `xtz` so I followed the same structure, now they are deduped in baseCoin/iface.ts

src/coin/eth/keyPair.ts
gonzaalvarez70: Sure, I've just refactored the keyPair files to use a baseKeyPair class, I'll upload it soon.

src/coin/eth/signature.ts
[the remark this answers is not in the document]
argjv: @marktoda is right. I found we used these internally: 

```
"ethereumjs-abi": "0.6.0",
"ethereumjs-tx": "1.1.1",
"ethereumjs-util": "4.4.0",
```

If we stick to the same version, we can use them without a thorough audit. However, try to use them only when it is absolutely required.

src/coin/eth/transaction.ts
gonzaalvarez70: Nice catch, it was an auto-import, thanks!

src/coin/eth/transactionBuilder.ts
gonzaalvarez70: Yes, that's a better way to do it.

src/coin/eth/utils.ts
gonzaalvarez70: We need to send the arguments to the wallet simple constructor, that's why we need to do this. But we are working on an alternative to avoid the web3 usage.

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /** @inheritdoc */
+  protected fromImplementation(rawTransaction: string): Transaction {
+    // Decoding the transaction is an async operation, so save it and leave the decoding for the
+    // build step
+    throw new BuildTransactionError('Not implemented fromImplementation'); //TODO: Implement fromImplementation and delete this
+  }
+
+  protected signImplementation(key: BaseKey): BaseTransaction {
```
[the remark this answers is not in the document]
gonzaalvarez70: Yes, that's the intention, I will add the same comment as in tezos stating: 
```
    // Signing the transaction is an async operation, so save the source and leave the actual
    // signing for the build step
```

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  // region Common builder methods
+  /**
+   * Set the transaction chain id.
+   *
+   * @param {number} chainId A block hash to use as branch reference
+   */
+  chainId(chainId: number): void {
+    this._chainId = chainId;
```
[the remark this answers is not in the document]
gonzaalvarez70: We are planning to add that to a future revision. For now, the intention is to keep it that way.

src/coin/eth/utils.ts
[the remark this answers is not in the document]
gonzaalvarez70: Y finally managed to get it working using `'0x' + Number(transactionData.gasLimit as number).toString(16)`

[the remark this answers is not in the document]
marktoda: I was more concerned about the `as number` casting -- why is that necessary?

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,216 @@
...
+    this._walletOwnerAddresses.push(address);
+  }
+  //endregion
+
+  /**
+   * Build a transaction for a generic multisig contract.
+   *
+   * @returns {TxData} The Ethereum transaction data
+   */
+  private buildWalletInitializationTransaction(): TxData {
```
argjv, line 197: This goes into the `region WalletInitialization builder methods` region above

package.json
```diff
@@ -42,7 +42,10 @@
     "elliptic": "^6.5.2",
     "libsodium-wrappers": "^0.7.6",
     "protobufjs": "^6.8.9",
-    "tronweb": "^2.7.2"
+    "tronweb": "^2.7.2",
+    "keccak": "^1.0.2",
+    "rlp": "^2.0.0",
```
argjv, line 47: Doesn't look you need this anymore.
You have to also update `package-lock.json`

src/coin/baseCoin/baseKeyPair.ts
```diff
@@ -0,0 +1,84 @@
+import * as Crypto from '../../utils/crypto';
+import { KeyPairOptions, ExtendedKeys } from './iface';
+import { AddressFormat } from './enum';
+import { HDNode, ECPair } from 'bitgo-utxo-lib';
+
+/**
+ * Base keys and address management.
+ */
+export abstract class BaseKeyPair {
```
argjv, line 9: s/`BaseKeyPair`/`ExtendedKeyPair`/g

src/coin/eth/iface.ts
argjv: If you remove the interfaces like Mark suggests, you may be able to.

src/coin/eth/keyPair.ts
```diff
@@ -0,0 +1,67 @@
...
+   * Public constructor. By default, creates a key pair with a random master seed.
+   *
+   * @param { KeyPairOptions } source Either a master seed, a private key (extended or raw), or a public key
+   *     (extended, compressed, or uncompressed)
+   */
+  constructor(source?: KeyPairOptions) {
+    super(source);
+    if (source) {
+      if (isPrivateKey(source)) {
+        super.recordKeysFromPrivateKey(source.prv);
```
argjv, line 20: Instead of a base class, you can just instantiate the class like and use these methods. That way we `prefer composition over inheritance`

src/coin/baseCoin/baseKeyPair.ts
gonzaalvarez70: Sorry, I didn't get this, what do you mean with this comment?

src/coin/eth/transactionBuilder.ts
FabianSimtlix: done

src/coin/eth/transactionBuilder.ts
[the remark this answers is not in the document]
FabianSimtlix: I added a TODO to make this change

src/coin/eth/keyPair.ts
silalvado: BaseKeyPar  is an abstract class and cannot be instantiated.  If we don't misunderstand,  it is designed to be inherited.

src/coin/eth/keyPair.ts
silalvado: Done.

src/coin/baseCoin/baseKeyPair.ts
argjv: Rename `BaseKeyPair` to `ExtendedKeyPair`.
It's from `sed` [command](https://unix.stackexchange.com/questions/484659/what-does-this-command-do-sed-s-g)

package.json
argjv: I don't see it being used directly: https://github.com/paszyrko/bitgo-account-lib/search?q=rlp&unscoped_q=rlp
I thought it was part of the signature code copied one of eth libs

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,217 @@
...
+    }
+    // Signing the transaction is an async operation, so save the source and leave the actual
+    // signing for the build step
+    this._sourceKeyPair = signer;
+    return this.transaction;
+  }
+
+  /** @inheritdoc */
+  validateAddress(address: BaseAddress): void {
+    if (!isValidAddress(address.address)) {
```
argjv, line 91: We usually implement `isValidAddress(...)` in the coins utils function so that we can use it in other systems. Please wrap this into a method in `eth/utils.ts` (i.e. [Tezos](https://github.com/BitGo/bitgo-account-lib/blob/f09b5f38c4dab01e17e5f367009dfd51cc9f2ba4/src/coin/xtz/utils.ts#L179))

src/coin/eth/utils.ts
argjv: We need support for both wallet creation and transfer. Since this PR only supports wallet creation and not transfers (yet) it should be ok.
However, this `sign` method should evolve to support both transactions in the future.

test/unit/coin/eth/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  'f84b010a822ee080800162a03e02aa98cc658c89ce3f33b88cf70ac0f6294d89d9224483c9cd98a26853c932a02d29a3a6c42167d3ee92dc9dd490baada4074add90b10e8b390a13ac0f4e2ac5';
+
+describe('correct sign', function() {
+  it('should return a correct signed transaction', async () => {
+    const SIGNATURE = await sign(TXDATA, KEYPAIR);
+    should.equal(SIGNATURE, SIGN);
+  });
+});
+
+describe('Obtain contract data', function() {
```
argjv, line 25: Group the different `describe` into a top level one following the pattern of other tests files.

src/coin/eth/transactionBuilder.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /** @inheritdoc */
+  protected fromImplementation(rawTransaction: string): Transaction {
+    // Decoding the transaction is an async operation, so save it and leave the decoding for the
+    // build step
+    throw new BuildTransactionError('Not implemented fromImplementation'); //TODO: Implement fromImplementation and delete this
+  }
+
+  protected signImplementation(key: BaseKey): BaseTransaction {
```
marktoda, line 71: inheritDoc?

src/coin/eth/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Format transaction to be signed
+ *
+ * @param {TxData} transactionData the transaction data with base values
+ * @returns {TxData} the transaction data with hex values
+ */
+function formatTransaction(transactionData: TxData): TxData {
+  return {
+    gasLimit: '0x' + Number(transactionData.gasLimit as number).toString(16),
```
marktoda, line 45: I'm really not sure that this `as number` will always work, because by definition it's not always a number. I think we need good coverage of this to be sure it works. Even better, I'd prefer we found a way to avoid the casting altogether

src/coin/eth/utils.ts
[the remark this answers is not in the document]
gonzaalvarez70: You are right is not needed anymore. Removed in next commit.

package.json
gonzaalvarez70: Right, those dependencies are not needed. We will remove them.

src/coin/baseCoin/baseKeyPair.ts
gonzaalvarez70: Got it, file renamed.

src/coin/eth/transactionBuilder.ts
gonzaalvarez70: Done, library wrapped in a utils method.

test/unit/coin/eth/utils.ts
gonzaalvarez70: Added a group "ETH util library"

src/coin/eth/utils.ts
gonzaalvarez70: Done, removed this casting.

src/coin/baseCoin/iface.ts
```diff
@@ -2,6 +2,50 @@ export interface BaseKey {
...
+  return (source as PrivateKey).prv !== undefined;
+}
+
+/**
+ * @param source
+ */
+export function isPublicKey(source: KeyPairOptions): source is PublicKey {
+  return (source as PublicKey).pub !== undefined;
+}
+
```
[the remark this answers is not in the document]
gonzaalvarez70: Could we add this as a technical improvement to be developed next sprint? I think that this was not affected by the scope of the story. This way we can start working in parallel on other features.

src/coin/eth/transactionBuilder.ts
gonzaalvarez70: Done, added.

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export interface FieldStruct {
+  components?: any;
+  name: string;
+  inputs?: any;
+  type: string;
+}
+
+export interface Decoded {
```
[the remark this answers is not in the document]
gonzaalvarez70: Removed, thanks!

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,35 @@
...
+  data: Buffer | Buffer[];
+  remainder: Buffer;
+}
+
+export interface FieldData {
```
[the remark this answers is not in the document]
gonzaalvarez70: Removed, thanks!

src/coin/eth/abiCoder.ts
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Should be used to encode list of params
+ *
+ * @param {Array<string|object>} types the types to encode
+ * @param {Array<any>} params the params to be encoded
+ *
+ * @returns {string} encoded list of params
+ */
+export function encodeParameters(types: FieldStruct[], params: any): string {
```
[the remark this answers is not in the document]
gonzaalvarez70: A direct replacement didn't work but I agree that the usage of that library would be the best choice. We need to do a small research to make it work, though. I'd suggest adding this as a JIRA ticket and working on it separately.

src/coin/eth/iface.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+/**
+ * A transaction's data.
+ */
+export interface TxData extends ETHTxData {
+  /**
+   * The chainId's nonce.
+   */
+  chainId?: BufferLike;
```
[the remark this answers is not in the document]
gonzaalvarez70: We are not currently using ethereumjs-vm but we could evaluate to use it. However for Celo we need direct access to set the chainId on the transaction object being sent. I'd suggest leaving this as it is for now, and once the implementation for Celo is started checking if we can replace this for ethereumjs-vm.

src/coin/eth/abiCoder.ts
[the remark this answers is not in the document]
gonzaalvarez70: I managed to get it working, will upload this on the next commit. It should improve the coverage too.

src/coin/eth/iface.ts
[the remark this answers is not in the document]
argjv: This will be addressed in a follow up PR. If these are static values, we can just hardcode them as defaults somewhere, no need for a library.

src/coin/baseCoin/iface.ts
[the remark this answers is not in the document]
marktoda: sure

src/coin/eth/iface.ts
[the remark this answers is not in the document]
marktoda: `ethereumjs-vm` is just the repo that contains `ethereumjs-tx` which you are using. but OK

gonzaalvarez70:
We created a new PR on https://github.com/BitGo/bitgo-account-lib/pull/90
with each commit signed.

argjv:
Continued in https://github.com/BitGo/bitgo-account-lib/pull/90 to fix the commits signatures.

PR #91, opened 2020-04-29
Celo Wallet initialization

gonzaalvarez70:
Added subcoin behavior to support each coin, added Celo wallet initalization support.

 Ticket: BGX-45


src/coin/eth/transaction.ts
```diff
@@ -1,17 +1,18 @@
...
 import { InvalidTransactionError } from '../baseCoin/errors';
 import { KeyPair } from './keyPair';
 import { TxData } from './iface';
 import { Utils } from './';
-import { BaseCoin as CoinConfig } from '@bitgo/statics';
 
 export class Transaction extends BaseTransaction {
   private _parsedTransaction?: TxData; // transaction in JSON format
   private _encodedTransaction?: string; // transaction in hex format
+  private _subCoin: string;
```
argjv, line 15: This can evolve in complicated and confusing scenarios in the future (we've suffered it). We should have specific classes for each coin.
RSK and Celo must have their own `TransactionBuilder` and `Transaction` classes where we can put coin specific code, even if it is minimum or just a wrapper.
They can extend from the same Ethereum base class.

src/coin/eth/transaction.ts
```diff
@@ -59,15 +69,15 @@ export class Transaction extends BaseTransaction {
     if (!this._parsedTransaction) {
       throw new InvalidTransactionError('Empty transaction');
     }
-    this._encodedTransaction = await Utils.sign(this._parsedTransaction, keyPair);
+    this._encodedTransaction = await Utils.sign(this._parsedTransaction, this._subCoin, keyPair);
```
argjv, line 72: The Celo `Transaction` class should have it's own sign method and do something like: `this._encodedTransaction = await Utils.sign(this._parsedTransaction, this._coinConfig, keyPair);`

src/coin/eth/utils.ts
```diff
@@ -2,39 +2,99 @@ import { Buffer } from 'buffer';
...
+  return rawTransaction.raw;
+}
+
+/**
+ * Format transaction to be signed
+ *
+ * @param {TxData} transactionData the transaction data with base values
+ * @returns {TxData} the transaction data with hex values
+ */
+function formatCeloTx(transactionData: TxData): TxData {
```
argjv, line 89: This should live in `src/coin/celo/utils.ts`

src/coin/eth/utils.ts
```diff
@@ -2,39 +2,99 @@ import { Buffer } from 'buffer';
...
 }
 
+/**
+ * Signs a Celo transaction using celo contract kit
+ *
+ * @param {TxData} transactionData the transaction data to sign
+ * @param {KeyPair} keyPair the signer's keypair
+ * @returns {string} the transaction signed and encoded
+ */
+async function celoSign(transactionData: TxData, keyPair: KeyPair): Promise<string> {
```
argjv, line 76: This should live in `src/coin/celo/utils.ts`

src/index.ts
```diff
@@ -39,26 +39,31 @@ export const supportedCoins = Object.keys(coinBuilderMap);
...
 }
 
 /**
- * Get a transaction builder for the given coin.
+ * Maps the coin to eth if compatible, returns the same coin
+ * if it isn't
  *
  * @param {string} coin The coin name
- * @returns {string} The same coin or eth if it's compatible
+ * @returns {[string, string]} The coin type and the subcoin type
```
argjv, line 58: Each coin should have its own builder. Even if it is empty (which is not because at least the `_coinConfig` has to be set).

src/coin/eth/transaction.ts
twtaylorbitgo: We should be using inheritance to express this - this is not great design and will bite us down the road

gonzaalvarez70: Yes, sounds better. I will be working on that today.

src/coin/eth/transaction.ts
gonzaalvarez70: This was helpful, it's implemented this way now. Please check it.

src/coin/eth/utils.ts
gonzaalvarez70: Changed, thanks.

src/coin/eth/utils.ts
gonzaalvarez70: Changed, thanks.

src/index.ts
gonzaalvarez70: This is changed now. Thanks for the comment.

gonzaalvarez70:
> Looking at this PR - this needs better, more extensible design considerations before code is put to commit. A good design should be clear and concise
> 
> Inheriting from a subcoin and named functions based on the chain is very poor

@twtaylorbitgo You are right, it was my mistake, it's now fixed with the latest commit following the comments recommendation.

src/coin/eth/walletUtil.ts
```diff
@@ -139,3 +141,13 @@ export const walletSimpleAbi = [
     type: 'event',
   },
 ];
+
+export const customCommon = EthereumCommon.forCustomChain(
```
argjv, line 145: JSDoc and maybe change name? It is not clear what `customCommon` means. Is it a `KovanNetwork` or `EthTestnetConfig`?

test/unit/coin/rsk/transactionBuilder.ts
```diff
@@ -1,7 +1,8 @@
-import { getBuilder } from '../../../../src';
 import should from 'should';
+import { getBuilder } from '../../../../src';
 
 describe('RSK transaction builder should return an ETH Transaction builder', function() {
-  const txBuilder: any = getBuilder('rsk');
-  txBuilder.toString().should.equal(getBuilder('eth').toString());
+  //TODO: Implement rsk test with valid values.
```
argjv, line 5: Can't we do this now? I thought we were able to create valid wallet initialization transactions already

test/unit/coin/eth/transactionBuilder.ts
```diff
@@ -3,12 +3,13 @@ import { TransactionType } from '../../../../src/coin/baseCoin/';
...
...000000000000000000000000000000000000000000000000000000028152600401808373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200182815260200192505050602060405180830381600087803b151561040f57600080fd5b6102c65a03f1151561042057600080fd5b50505060405180519050151561043557600080fd5b5b50505050565b6000809054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff166108fc3073ffffffffffffffffffffffffffffffffffffffff16319081150290604051600060405180830381858888f1935050505015156104b457600080fd5b5600a165627a7a72305820088cce905e424bfe474510efe41761893e9e34b54b885b8fee546d8be88aee3c0029a165627a7a72305820827adc3ef3a179757bc33003e41070fbbfcdfc5db31275af3cb0395155cd3bbb0029000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000030000000000000000000000006461ec4e9db87cfe2aeec7d9b02aa264edfbf41f000000000000000000000000f10c8f42bd63d0aed3338a6b2b661bc6d9fa7c44000000000000000000000000a4b5666fb4ffea84dd848845e1114b84146de4b377a03e94c6e09e961f87fd8615233b99ecaa83b2106c486ffd66b3f72b5b3db39778a02c7706211bedfa602669169838342853bafbe29ba5a8a04051f4cd4e49049682';
```
argjv, line 6: Move these big strings into here: https://github.com/BitGo/bitgo-account-lib/tree/master/test/resources

test/unit/coin/etc/transactionBuilder.ts
```diff
@@ -1,7 +1,8 @@
-import { getBuilder } from '../../../../src';
 import should from 'should';
+import { getBuilder } from '../../../../src';
 
 describe('ETC transaction builder should return an ETH Transaction builder', function() {
-  const txBuilder: any = getBuilder('etc');
-  txBuilder.toString().should.equal(getBuilder('eth').toString());
+  //TODO: Implement ETC test with valid values
```
argjv, line 5: I'm adding ETC into statics too. Can you add a test for it once we have it there?

test/unit/coin/rsk/transactionBuilder.ts
gonzaalvarez70: Yes, I am doing that now, I just wanted to upload the rest of the code for feedback.

src/coin/eth/walletUtil.ts
gonzaalvarez70: Sure, I was thinking how to refactor this part. These are the customCommon attributes to initialize an ethereumtx object. The important part of that is the chainId. I am creating a new logic to get that from a resources file depending if it's mainnet or testnet, and depending on the coin.

test/unit/coin/eth/transactionBuilder.ts
gonzaalvarez70: Silvia is doing that on her commit. I'll wait until merging that to move each big string to avoid conflicts.

test/unit/coin/etc/transactionBuilder.ts
gonzaalvarez70: Sure, I'll add it to the mapping to.

src/coin/eth/utils.ts
```diff
@@ -7,32 +7,41 @@ import { SigningError } from '../baseCoin/errors';
...
-      chainId: transactionData.chainId as number,
-    },
-    'petersburg',
-  );
   const ethTx = new Transaction(formatTransaction(transactionData), { common: customCommon });
   const privateKey = Buffer.from(keyPair.getKeys().prv as string, 'hex');
   ethTx.sign(privateKey);
   const encodedTransaction = ethTx.serialize().toString('hex');
-  return encodedTransaction;
+  return '0x' + encodedTransaction;
```
marktoda, line 33: nit: prefer `ethUtil.addHexPrefix` or a similar helper function to ensure you never end up with `0x0x..` -- that's an easy bug to write, and easy to avoid with a helper

gonzaalvarez70: Sounds good! Thanks

gonzaalvarez70:
This PR has now conflicts with the code from https://github.com/BitGo/bitgo-account-lib/pull/92
I'll solve them and submit the changes tomorrow. But the code will be pretty much the same.

src/coin/cgld/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  return rawTransaction.raw;
+}
+
+/**
+ * Format transaction to be signed
+ *
+ * @param {TxData} transactionData the transaction data with base values
+ * @returns {TxData} the transaction data with hex values
+ */
+function formatTx(transactionData: TxData): TxData {
```
marktoda, line 29: What is the point of this function? It takes something of type `TxData` and returns something of type `TxData` but wrangles the types of the fields inside? I'm confused, it seems like there should be 2 different types here. At least explain it better in the comment

src/coin/cgld/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  return {
+    nonce: addHexPrefix(Number(transactionData.nonce).toString(16)),
+    data: transactionData.data,
+    gasLimit: addHexPrefix(Number(transactionData.gasLimit).toString(16)),
+    gasPrice: addHexPrefix(Number(transactionData.gasPrice).toString(16)),
```
marktoda, line 34: I think gasPrice as number is potentially lossy as it's a 256 bit int

src/coin/eth/resources.ts
```diff
@@ -0,0 +1,27 @@
...
+  'petersburg',
+);
+
+/**
+ * A Common object defining the chain and the hardfork for Ethereum Mainnet
+ */
+export const mainnetCommon = EthereumCommon.forCustomChain(
+  'mainnet',
+  {
+    name: 'testnet',
```
marktoda, line 22: ? mainnet?

src/coin/eth/utils.ts
```diff
@@ -43,9 +52,9 @@ export async function sign(transactionData: TxData, keyPair: KeyPair): Promise<s
  */
 function formatTransaction(transactionData: TxData): TxData {
```
marktoda, line 53: Same comment as above, these functions are really confusing to me and it seems that there should be 2 types -- what is the input type for, and where do we get it from? What is the output type for, and who consumes it?

src/coin/rbtc/resources.ts
```diff
@@ -0,0 +1,27 @@
+import EthereumCommon from 'ethereumjs-common';
+
+/**
+ * A Common object defining the chain and the hardfork for RSK Testnet
+ */
+export const testnetCommon = EthereumCommon.forCustomChain(
+  'ropsten',
+  {
```
marktoda, line 8: what is the difference between `networkId` and `chainId`? They are the same in all of your definitions like this

src/coin/cgld/utils.ts
gonzaalvarez70: That's right it's formatting the transaction using hex values for each fields as it needed for the libraries. Do you think that we should create a new type? I can add more info to the JsDoc.

src/coin/cgld/utils.ts
gonzaalvarez70: We should use something like BigNumber instead?

src/coin/eth/resources.ts
gonzaalvarez70: My bad, this one would be mainnet

src/coin/rbtc/resources.ts
gonzaalvarez70: For ETC they are different. For most networks, both parameters are the same, but there are some that don't match.

src/coin/eth/utils.ts
gonzaalvarez70: This TxData type is inherited from ethereumjs-tx. It has defined the types as BufferLike = Buffer | TransformableToBuffer | PrefixedHexString | number
Do you think that we should save the value as Hex directly when we receive the input? So we can avoid this formatting.

src/coin/rbtc/resources.ts
marktoda: :+1:

src/coin/cgld/utils.ts
marktoda: :+1:

package.json
```diff
@@ -32,7 +32,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@bitgo/statics": "^3.4.2",
+    "@bitgo/statics": "http://s3-us-west-2.amazonaws.com/indexer-status.dev.bitgo.com/statics.tar.gz",
```
marktoda, line 35: note this is just a stopgap hack until we can upload to npm :D

PR #92, opened 2020-04-29
Add validations in ETH transaction builder

silalvado:
It was added validations for transactions, raw transaction and keys


test/unit/coin/eth/transactionBuilder.ts
```diff
@@ -176,5 +234,42 @@ describe('Eth Transaction builder', function() {
         'A maximum of 3 owners can be set for a multisig wallet',
       );
     });
+    it('should be a valid address', () => {
```
argjv, line 237: nit: add a new line before `it()`

test/unit/coin/eth/transactionBuilder.ts
```diff
@@ -1,9 +1,13 @@
 import should from 'should';
 import { TransactionType } from '../../../../src/coin/baseCoin/';
 import { getBuilder, Eth } from '../../../../src';
-
+const TX_JSON =
```
argjv, line 4: Can we put these big strings in the test resource folder?

src/coin/eth/transactionBuilder.ts
```diff
@@ -62,9 +69,14 @@ export class TransactionBuilder extends BaseTransactionBuilder {
 
   /** @inheritdoc */
   protected fromImplementation(rawTransaction: string): Transaction {
-    // Decoding the transaction is an async operation, so save it and leave the decoding for the
-    // build step
-    throw new BuildTransactionError('Not implemented fromImplementation'); //TODO: Implement fromImplementation and delete this
+    const tx = new Transaction(this._coinConfig);
+    if (/^0x?[0-9a-f]{1,}$/.test(rawTransaction.toLowerCase())) {
+      tx.setEncodedTransactionData(rawTransaction);
```
argjv, line 74: These 2 methods could be constructors

test/unit/coin/eth/transactionBuilder.ts
silalvado: Done.

test/unit/coin/eth/transactionBuilder.ts
silalvado: Done.

src/coin/eth/transactionBuilder.ts
silalvado: I added those methods to the constructor, with the data as an optional parameter. However, I didn't eliminate the methods to be able set this data after construction.

argjv: I'm not sure that method is needed, do you have an example?
So far we only pass a hex encoded transaction or the TxData when we need to parse and then sign the transaction.
This method suggests that you can build a transaction and then override it with an existing one, which is odd.

silalvado: I think you are right, thank you. I already removed this method.

PR #93, opened 2020-05-09
Celo wallet initialization from serialized transaction

silalvado:
This PR includes deserialization of Celo wallet initialization transactions.


src/coin/eth/transaction.ts
```diff
@@ -82,4 +84,27 @@ export class Transaction extends BaseTransaction {
...
+  initFromSerializedTransaction(serializedTransaction: string): void {
+    this._encodedTransaction = serializedTransaction;
+    const decodedTx = RLP.decode(serializedTransaction);
+    const [rawNonce, rawGasPrice, rawGasLimit, rawTo, rawValue, rawData, rawV, rawR, rawS] = decodedTx;
+    const parsedTransaction: TxData = {
+      nonce: bigNumberify(rawNonce).toNumber(),
+      gasPrice: bigNumberify(rawGasPrice).toNumber(),
+      gasLimit: bigNumberify(rawGasLimit).toNumber(),
+      to: rawTo,
+      value: bigNumberify(rawValue).toNumber(),
```
marktoda, line 102: This is lossy as eth values are 256 bits.

src/coin/eth/transactionBuilder.ts
```diff
@@ -69,13 +72,9 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
   /** @inheritdoc */
   protected fromImplementation(rawTransaction: string): Transaction {
-    let tx: Transaction;
-    if (/^0x?[0-9a-f]{1,}$/.test(rawTransaction.toLowerCase())) {
-      tx = new Transaction(this._coinConfig, rawTransaction);
-    } else {
-      const txData = JSON.parse(rawTransaction);
-      tx = new Transaction(this._coinConfig, txData);
-    }
+    //rawTransaction is already validated in validateRawTransaction method
```
marktoda, line 75: I think it would be better if you did the deserialization work here, and constructed the transaction with `TxData` object as usual. Adding more different ways of storing the same data in `Transaction` class is making the code there quite complicated as we have to continuously check which type of storage model (encoded string or decoded txData object) we are storing..

marktoda: Alternatively maybe a static constructor function like `fromSerialized` would be good

gonzaalvarez70: We are using the _serializedTransaction value to check if the transaction building begun with a from(). This was done based on the Tezos implementation, so if there was a serialized transaction present, then the transaction is being built from a serialized tx. For example when we do a validate transaction, we don't have the needed fields (i.e. transactionBuilder._walletOwnerAddresses). Do you have any other idea to determine this? Or we should keep that field?

src/coin/eth/transaction.ts
silalvado: You're right. I changed it to a big number in a string format.

src/coin/eth/transactionBuilder.ts
marktoda: To give you a better idea of what I'm talking about, I implemented it here: https://github.com/BitGo/bitgo-account-lib/pull/95 -- please take a look. We can either merge that then fix this on top, or you guys can take my ideas into account here

silalvado: I changed deserialization to transaction builder. However I kept  _serializedTransaction because we need it to identify if the transaction building begun with a from().

marktoda: It also closes BGA-56 which was marked "done" though never actually completed.

marktoda: I disagree, you don't need to differentiate between transactions that begun with a `from` or not -- they are the same thing! See my example implementation: https://github.com/BitGo/bitgo-account-lib/pull/95/files

gonzaalvarez70: One issue that I see is with the wallet owners values, they are encoded in the "data" field. How would that work if we don't use that. And with sendMultiSig for example we wouldn't have all of the internal values either.

marktoda: You can deserialize the data field -- in fact you will need to do this to perform proper validation of serialized transactions anyways. Better to do it in one central location, so you can reuse the code

src/coin/cgld/transactionBuilder.ts
```diff
@@ -1,10 +1,72 @@
...
+      value: bigNumberify(rawValue).toString(),
+      data: rawData,
+      chainId: 44786,
+      v: rawV,
+      r: rawR,
+      s: rawS,
+    };
+    const tx = this.transaction;
+    tx.setTransactionData(parsedTransaction);
+    return tx;
```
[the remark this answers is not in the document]
marktoda: ah, forgot this was CELO -- nevermind. Where did you get this deserialization logic from?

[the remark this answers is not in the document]
silalvado: We made this based on the celo library, If you want, we can try to use the library method. I don't know what is best for you.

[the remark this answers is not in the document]
marktoda: Ah OK -- let's just use the library method!

gonzaalvarez70:
This PR was deprecated in favor of https://github.com/BitGo/bitgo-account-lib/pull/101
Story BGA-10 work will be finished there.

PR #94, opened 2020-05-12
Fix Transactionbuilder and keypair

marktoda:
The transactionbuilder and keypair for ETH were broken with the
following bugs:
- keypair would not generate addresses from an extended key, simply due
to not handling the case
- transactionbuilder would throw if not given a counter, even though
there is a default

This commit fixes the above issues and adds some tests

Ticket: BG-20990

### DIFF
--- a/src/coin/eth/transaction.ts
+++ b/src/coin/eth/transaction.ts
@@ -12,7 +12,6 @@ import { Utils } from './';
 export class Transaction extends BaseTransaction {
   protected _parsedTransaction?: TxData; // transaction in JSON format
   protected _encodedTransaction?: string; // transaction in hex format
-  private _source: string;
 
   /**
    * Public constructor.
--- a/src/coin/eth/transactionBuilder.ts
+++ b/src/coin/eth/transactionBuilder.ts
@@ -21,7 +21,6 @@ const DEFAULT_M = 3;
  * Ethereum transaction builder.
  */
 export class TransactionBuilder extends BaseTransactionBuilder {
-  private _serializedTransaction: string;
   private _transaction: Transaction;
   private _sourceKeyPair: KeyPair;
   private _type: TransactionType;
@@ -31,7 +30,6 @@ export class TransactionBuilder extends BaseTransactionBuilder {
   private _sourceAddress: string;
 
   // Wallet initialization transaction parameters
-  private _initialBalance: string;
   private _walletOwnerAddresses: string[];
 
   /**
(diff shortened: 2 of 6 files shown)


src/coin/eth/transactionBuilder.ts
```diff
@@ -143,20 +141,36 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+        }
+
+        if (this._walletOwnerAddresses.length !== 3) {
+          throw new BuildTransactionError(
+            `Invalid transaction: wrong number of owners -- required: 3, ` +
+              `found: ${this._walletOwnerAddresses.length}`,
+          );
+        }
+
+        if (this._counter === undefined) {
```
argjv, line 165: Is this the nonce? Counter is the Tezos name for nonces.
Also, if it is a string, isn't it better to check for `if (!this._counter)`?

src/coin/eth/transactionBuilder.ts
```diff
@@ -143,20 +141,36 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+          throw new BuildTransactionError('Invalid transaction: missing address counter');
+        }
+
+        if (this._sourceAddress === undefined) {
```
argjv, line 169: Ditto.

src/coin/eth/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { Buffer } from 'buffer';
+import { HDNode } from 'bitgo-utxo-lib';
```
argjv, line 1: Why is this change needed?

src/coin/eth/transaction.ts
```diff
@@ -2,6 +2,7 @@
  * Ethereum transaction model
  */
 import { BaseCoin as CoinConfig } from '@bitgo/statics';
+import transaction from 'ethereumjs-tx/dist/transaction';
```
argjv, line 5: Is this needed? It is not used in this PR

src/coin/eth/transactionBuilder.ts
marktoda: no, it's a number and it can be 0. that is one of the bugs I'm fixing here

src/coin/eth/utils.ts
marktoda: we use xpubs for ETH, previously it would throw if given an xpub

src/coin/eth/transaction.ts
marktoda: removed

src/coin/eth/utils.ts
marktoda: removed

src/coin/eth/transactionBuilder.ts
marktoda: I'd prefer the name is nonce myself, but I don't want to step over simtlix team's toes as they probably have other PRs built on that name, and it's fine for consistency with xtz. I just wanted to fix the bug to unblock myself

marktoda: for context, I tried to send a tx with nonce 0 (First tx on the wallet) and it failed because of the `(!this._counter)` check.

src/coin/eth/transactionBuilder.ts
marktoda: fixed!

PR #96, opened 2020-05-13
Add test for ETH wallet initialization ser/deser

marktoda:
This commit adds a test that ensures that we can serialize and
deserialize transactions back-and-forth and maintain the same output.
This is required by the internal bitgo components as they pass around
serialized transactions and need to be able to parse them again.

Ticket: BG-21074


marktoda:
fixed

PR #98, opened 2020-05-14
Bga 11 send multisig new structure

FabianSimtlix:
Add sendMultiSig contract method call for celo


src/coin/eth/transactionBuilder.ts
```diff
@@ -161,19 +175,13 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
-
         break;
       case TransactionType.Send:
-        //  TODO: validate when transaction type send be developed
-        throw new BuildTransactionError('Unsupported transaction type');
+        this.validateBaseTransactionFields();
+        if (this._contractAddress === undefined) {
+          throw new BuildTransactionError('Invalid transaction: missing contract address');
+        }
+        break;
```
marktoda, line 184: Don't we also need:
- value
- sequence id
- expiration
?

marktoda: Ah I see now that these are encapsulated in `_transfer` but we should make sure that that is defined and has everything it needs to `build`, right?

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+    this._amount = amount;
+    return this;
+  }
+
+  to(address: string): TransferBuilder {
+    this._toAddress = address;
+    return this;
+  }
+
+  sequenceId(counter: number): TransferBuilder {
```
marktoda, line 31: `contractSequenceId`

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+  private _signKey: string;
+  private _expirationTime: number;
+
+  //initialize with default values for non mandatory fields
+  constructor() {
+    this._expirationTime = this.getExpirationTime();
+    this._data = '0x';
+  }
+
+  amount(amount: number): TransferBuilder {
```
marktoda, line 21: amount can't be `number` -- it's 256 bits in ETH so this could be lossy. Please take a number string, and use BigNumber for arithmetic

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+        signature,
+      );
+    }
+    throw new BuildTransactionError(
+      'Missing transfer mandatory fields. Amount, destination (to) address, signing key and sequenceID are mandatory',
+    );
+  }
+
+  private hasMandatoryFields(): boolean {
+    return this._amount && this._toAddress && this._sequenceId && this._signKey ? true : false;
```
marktoda, line 69: Why `this._signKey ? true : false`? That should have the same effect as just `this._signKey`

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+    this._expirationTime = this.getExpirationTime();
+    this._data = '0x';
+  }
+
+  amount(amount: number): TransferBuilder {
+    this._amount = amount;
+    return this;
+  }
+
+  to(address: string): TransferBuilder {
```
marktoda, line 26: Maybe do some validation to make sure it is a valid address?

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+    this._sequenceId = counter;
+    return this;
+  }
+
+  data(additionalData: string): TransferBuilder {
+    this._data = additionalData;
+    return this;
+  }
+
+  key(signKey: string): TransferBuilder {
```
marktoda, line 41: I think this should take a `KeyPair` -right?

marktoda: Also do we need to take this as part of the builder? Maybe the `sign` function should just take the key as a parameter

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  expirationTime(date: number): TransferBuilder {
+    this._expirationTime = date;
+    return this;
+  }
+
+  build(): string {
+    if (this.hasMandatoryFields()) {
+      const signature = this.ethSignMsgHash();
```
marktoda, line 53: Ah I see the issue, we can't separate out build / sign because the signature is part of the data. Hmmmm OK -- can we change the name of `key` to `sign` then or change this function to be `buildAndSign` ro make it more clear that the signing step has already happened?

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+  private hasMandatoryFields(): boolean {
+    return this._amount && this._toAddress && this._sequenceId && this._signKey ? true : false;
+  }
+
+  /** Return an expiration time, in seconds, set to one hour from now
+   *
+   * @returns {number} expiration time
+   */
+  private getExpirationTime(): number {
+    const currentTime = new Date().getTime() / 1000;
```
marktoda, line 77: readability nit:  it might be more clear to do it like this:
```
const currentDate = new Date();
currentDate.setHours(currentDate.getHours() + 1);
return currentDate.getTime() / 1000;
```

Just to make it more clear that you're changing the hours by 1, so users don't have to figure out what the 3600 means.  Otherwise, please move the `3600` to a variable like `ONE_HOUR_IN_SECONDS` for readability

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
...
+      new Buffer(this._signKey, 'hex'),
+    );
+
+    // Assemble strings from r, s and v
+    const r = ethUtil.setLengthLeft(signatureInParts.r, 32).toString('hex');
+    const s = ethUtil.setLengthLeft(signatureInParts.s, 32).toString('hex');
+    const v = ethUtil.stripHexPrefix(ethUtil.intToHex(signatureInParts.v));
+
+    // Concatenate the r, s and v parts to make the signature string
+    return ethUtil.addHexPrefix(r.concat(s, v));
```
marktoda, line 108: We should decide if we want to store the signature as `r.concat(s, v)` string, or as `v, r, s` separately like you have in `TxData` interface. I don't really care which one, but we should be consistent. Maybe the former is easier so we can save them in the `signatures` array in `BaseTransaction`: https://github.com/BitGo/bitgo-account-lib/blob/0ffb8b7238beaea836bb8df011e2e65d6c535927/src/coin/baseCoin/baseTransaction.ts#L13

src/coin/eth/types.ts
```diff
@@ -14,17 +14,17 @@ export class EthTransaction {
...
-        nonce: tx.nonce,
+        nonce: addHexPrefix(new BigNumber(tx.nonce).toString(16)),
         to: tx.to,
         gasPrice: addHexPrefix(new BigNumber(tx.gasPrice).toString(16)),
         gasLimit: addHexPrefix(new BigNumber(tx.gasLimit).toString(16)),
         value: addHexPrefix(new BigNumber(tx.value).toString(16)),
         data: tx.data,
       }),
-      tx.chainId,
+      addHexPrefix(new BigNumber(Number(tx.chainId)).toString(16)),
```
marktoda, line 27: is chainId a string or a number? Let's just pick one so you don't have to cast here, then if we need the other we can convert it with an adapter

src/coin/eth/types.ts
```diff
@@ -14,17 +14,17 @@ export class EthTransaction {
    *
    * @param tx The JSON representation of the transaction
    */
-  public static fromJson(tx: TxJson): EthTransaction {
+  public static fromJson(tx: TxData): EthTransaction {
     return new EthTransaction(
       new EthereumTx({
-        nonce: tx.nonce,
+        nonce: addHexPrefix(new BigNumber(tx.nonce).toString(16)),
```
marktoda, line 20: Why is this needed? I think `EthereumTx` can handle nonce as a number, right? Or does it need to be a hex string?

src/coin/cgld/utils.ts
```diff
@@ -1,39 +1,41 @@
...
  *
- * @param {TxJson} transactionData the transaction data with base values
- * @returns {CeloTxData} the transaction data with hex values, ready for CELO signing library
+ * @param {TxData} transactionData the transaction data unfiltered
+ * @returns {TxData} the transaction data filtered with the available values
  */
-function formatTx(transactionData: TxJson): CeloTxData {
-  return {
-    nonce: addHexPrefix(Number(transactionData.nonce).toString(16)),
+function filterTx(transactionData: TxData): TxData {
```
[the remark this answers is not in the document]
FabianSimtlix: I rename it to removeEmptyFields to be more clear about what it does.
I think it's ok that we return the same type as what we are trying to accomplish here is to clear the empty fields as the contract kit library does not handle them.

src/coin/cgld/utils.ts
```diff
@@ -1,39 +1,41 @@
...
     data: transactionData.data,
-    gasLimit: addHexPrefix(Number(transactionData.gasLimit).toString(16)),
-    gasPrice: addHexPrefix(new BigNumber(transactionData.gasPrice as string).toString(16)),
-    chainId: addHexPrefix(Number(transactionData.chainId).toString(16)),
+    gasLimit: transactionData.gasLimit,
+    gasPrice: transactionData.gasPrice,
+    chainId: transactionData.chainId,
+    value: transactionData.value,
   };
+  if (transactionData.to && transactionData.to !== '0x') filtered.to = transactionData.to;
```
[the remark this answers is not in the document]
FabianSimtlix: done

src/coin/eth/iface.ts
```diff
@@ -4,20 +4,36 @@ export interface Fee extends BaseFee {
...
 
+/**
+ * A transaction's data.
+ */
+export interface TxData {
+  //TODO: Define each field type
+  gasLimit: string;
+  gasPrice: string;
+  to?: string;
+  nonce: string | number;
```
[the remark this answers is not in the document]
FabianSimtlix: we are not using compound types anymore

src/coin/eth/iface.ts
```diff
@@ -4,20 +4,36 @@ export interface Fee extends BaseFee {
   gasLimit: string;
 }
 
+/**
+ * A transaction's data.
+ */
+export interface TxData {
```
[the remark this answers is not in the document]
FabianSimtlix: I keep the TxData name to be consistent with the type name on EthereumTx. I agree with you that TxJson could have more meaning as what it is (a json representation of the transaction) but at the process level I think that TxData gives more context. But we can define one or the other

src/coin/eth/transferBuilder.ts
FabianSimtlix: done

src/coin/eth/transferBuilder.ts
FabianSimtlix: I agree, it's more clear that way .
Done

src/coin/eth/transferBuilder.ts
FabianSimtlix: this signature is specific to the method call arguments, not the transaction data. we are not going to store the result of this function.

src/coin/eth/types.ts
FabianSimtlix: it's an string, here we init the Transaction and set the value in its hex form as we need it as an hex value on the flows. So casting it here allows us to not have to manipulate the value anymore

src/coin/eth/transactionBuilder.ts
```diff
@@ -138,18 +143,27 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
       case TransactionType.WalletInitialization:
         // assume sanitization happened in the builder function, just check that all required fields are set
-        if (this._fee === undefined) {
-          throw new BuildTransactionError('Invalid transaction: missing fee');
-        }
-
-        if (this._chainId === undefined) {
-          throw new BuildTransactionError('Invalid transaction: missing chain id');
-        }
+        this.validateBaseTransactionFields();
```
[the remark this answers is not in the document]
FabianSimtlix: Done

src/coin/eth/transactionBuilder.ts
FabianSimtlix: The next step will perform that validation, so i think it's better to not put those values here. This way TransactionBuilder is unaware of contract method specific parameters, and just handle the transaction related values

src/coin/eth/transferBuilder.ts
FabianSimtlix: It's the sum off all conditions not just this._signKey. For what I saw force me to do be explicit and use ternary "? true : false"

src/coin/eth/transferBuilder.ts
FabianSimtlix: Done

src/coin/eth/transferBuilder.ts
FabianSimtlix: Done

src/coin/eth/transferBuilder.ts
FabianSimtlix: Done

src/coin/eth/iface.ts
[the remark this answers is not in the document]
marktoda: OK, no problem I'm OK keeping it

src/coin/eth/transactionBuilder.ts
marktoda: What next step? Can we call `this._transfer.validate()` or something?

src/coin/eth/transferBuilder.ts
marktoda: Hmmm OK -- it really makes you use the ternary? In that case, can you wrap the main part in parens so it's more clear? I think most people don't know the order of operations off the top of their head, so it's hard to know where the ternary applies

src/coin/eth/transferBuilder.ts
marktoda: I think it should take the same type that `TransactionBuilder.sign` takes, like see this call that I copied from the tests:

```
 txBuilder.sign({ key: defaultKeyPair.getKeys().prv });	
```

A call to `transferBuilder` should look the same, whatever that may be

src/coin/eth/transferBuilder.ts
marktoda: Why are you not going to store it? I'd imagine when you need to deserialize the sendMultisig transaction data in `from` you'll need a data structure to store the details of the method call

src/coin/eth/types.ts
marktoda: I think since you removed the compound type from `TxData` you can remove the cast.

src/coin/cgld/transaction.ts
```diff
@@ -1,9 +1,34 @@
...
+   * it or not.
+   *
+   * @param {KeyPair} keyPair The key to sign the transaction with
+   */
+  async sign(keyPair: KeyPair): Promise<void> {
+    // Check if there is a transaction to sign
+    if (!this._ethTransaction) {
+      throw new InvalidTransactionError('Empty transaction');
+    }
+    this._encodedTransaction = await Utils.sign(this._ethTransaction.toJson(), keyPair);
```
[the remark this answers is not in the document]
marktoda: What was your question here?

src/coin/eth/transferBuilder.ts
marktoda: Now that I see it, you need to check for `=== undefined` for amount and sequence id anyways since `0` and `""` are falsy

marktoda: So maybe this:
```js
return this._amount !== undefined && this._toAddress !== undefined && this._sequenceId !== undefined && this._signKey !== undefined
```

explicit, and correct

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,110 @@
+import ethUtil from 'ethereumjs-util';
+import EthereumAbi from 'ethereumjs-abi';
+import { BuildTransactionError } from '../baseCoin/errors';
+import { sendMultiSigData } from './utils';
+
+/** ETH transfer builder */
+export class TransferBuilder {
+  private _amount: number;
```
marktoda: lossy

src/coin/cgld/types.ts
```diff
@@ -0,0 +1,95 @@
+import BigNumber from 'bignumber.js';
+import { signTransaction } from '@celo/contractkit/lib/utils/signing-utils';
+import { addHexPrefix, toBuffer, bufferToHex, bufferToInt, rlp } from 'ethereumjs-util';
+import { TxData } from '../eth/iface';
+import { EthereumTransaction, EthTransaction } from '../eth/types';
+import { KeyPair } from '../eth';
+
+export class CeloTransaction extends EthereumTransaction {
```
marktoda, line 8: Why is both this and `CgldTransaction` below necessary? It seems that they only 1 is needed. Maybe it will help to use an interface for the `EthTransaction` instead so you aren't constrained to the field types used in the superclass

FabianSimtlix: Both are needed, because sign, serialize and de-serialize process are different.
EthereumTransaction is the interface that allow us tu join the two different logic from Ethereumjs-tx and celo contract-kit. and share the methods on EthTransaction and also on Transaction class.

FabianSimtlix: As we implement the other functionalities (BGA-10, BGA-62, BGA-12) we will continue cleaning this

src/coin/eth/transactionBuilder.ts
FabianSimtlix: next step in the process will attempt to create de transaction, and at that point transferBuilder will validate it has his values.
This is a coding style, personally I prefer this class to just handle the transaction related values.

src/coin/eth/transferBuilder.ts
FabianSimtlix: I think both are valid.
We can have a sign method here similar to transactionBuilder. but I think it shouldn't generate the signature until build. Otherwise the user will have to be aware that the sign is the last parameter that he pass to the builder or it will fail or produce an invalid signature value.
The other option (receiving the key on building) also will work. But that wouldn't be consistent with the build method on transactionBuilder or the transferBuilder implementation on other coins.

src/coin/eth/transferBuilder.ts
FabianSimtlix: I add field to store this result. I defined it as string, at least for now. What I meant to say before is that this value is not going to be saved in BaseTransaction signatures property. I think the scope of this  two signatures are very different

src/coin/eth/transferBuilder.ts
```diff
@@ -0,0 +1,121 @@
...
+   *
+   * @returns {number} expiration time
+   */
+  private getExpirationTime(): number {
+    const currentDate = new Date();
+    currentDate.setHours(currentDate.getHours() + 1);
+    return currentDate.getTime() / 1000;
+  }
+
+  private getSHA(): (string | Buffer)[][] {
```
marktoda, line 92: nit: Naming could be better -- you're not actually _getting the hash_, just getting the data that will be hashed. I'd suggest you either rename it to "getOperationData" then hash it later, or just change the function to include `ethUtil.bufferToHex(EthereumAbi.soliditySHA3(...this.getSHA()));`. In that case `getOperationHash` is a better name as well

src/coin/cgld/types.ts
marktoda: I still think that you can and should find a way to merge these two classes -- their names are very similar and confusing, and the difference between them is not clear. I also don't think that we should merge code with the intention of cleaning it up _later_ -- we should be constantly thinking of how to improve the quality of our code and working to improve it as we go along, not pushing it off to later

marktoda: Maybe the best way to merge them would be to create an interface: `EthTransaction` which exposes `fromJson`, `toJson`, `fromSerialized` and `toSerialized` -- then it would be very easy to have different data structures inside the classes and simply implement those four functions.

marktoda:
Approved if you really want to land this now, but I still have some unresolved comments above.

src/coin/cgld/types.ts
```diff
@@ -0,0 +1,160 @@
...
+      return this._senderPubKey;
+    }
+    throw new Error('Invalid Signature');
+  }
+
+  serialize(): Buffer {
+    return rlp.encode(this.raw);
+  }
+
+  sign(privateKey: Buffer): void {
```
marktoda, line 102: How does this work? This seems to require v, r, and s already existing when this class is instantiated. How will we sign for the first time, then? Also `this._signatures` containing v, r, s, and the private key doesn't make much sense, why is the private key appended? That could also be a bad security issue if a user of this library assumes that `signatures` are safe to show (they usually are) but there's actually a private key in plain text right there! You could potentialy use `toRpcSig` to get a string representation of the signature: https://github.com/ethereumjs/ethereumjs-util/blob/master/docs/modules/_signature_.md#const-torpcsig

src/coin/eth/transferBuilder.ts
FabianSimtlix: Changed.

src/coin/cgld/types.ts
FabianSimtlix: done

PR #99, opened 2020-05-15
Implement Tezos forwarder contract

CapnMigraine:
Ticket: BG-21066


src/coin/baseCoin/enum.ts
```diff
@@ -6,6 +6,8 @@ export enum TransactionType {
   WalletInitialization,
   // Initialize a singlesig address. i.e. in Tezos, addresses have to be revealed
   AddressInitialization,
+  // Deploy forwarder contract to be used as wallet address
+  ForwarderContractDeployment,
```
marktoda, line 10: Thoughts: To me it seems like in BitGo context this is more like "address initialization" than the above. I think it makes more sense to have this be called `AddressInitialization` and the "reveal" (XTZ-specific thing) be called `AddressReveal` or something. That way ETH-like coins can re-use AddressInitialization and have it mean the same thing as here

CapnMigraine: I just saw this comment after pushing a change that does something similar. I agree and I used `PublicKeyRevelation` instead. It's very specific to Tezos, but I it's also more accurate and descriptive.

src/coin/xtz/multisigUtils.ts
```diff
@@ -57,6 +58,35 @@ export function getMultisigTransferDataFromOperation(operation: TransactionOp):
...
+    storageLimit: operation.storage_limit,
+  };
+
+  // const from = operation.source;
+  // const counter = operation.counter;
+  // const balance = operation.balance;
+  const { source: from, counter, balance } = operation;
+
+  // Transactions initializing a forwarder contract contain the address of the wallet contract
+  const contractAddress: string = _.get(operation, 'script.code[2].args[0][3].args[0][0].args[1].string');
```
marktoda, line 80: Is this seriously the best way to get the address? Such a crazy object format :laughing:

src/coin/baseCoin/enum.ts
```diff
@@ -3,9 +3,12 @@
  */
 export enum TransactionType {
   Send,
+  // Multisig contract deployment tx
   WalletInitialization,
-  // Initialize a singlesig address. i.e. in Tezos, addresses have to be revealed
+  // Forwarder contract deployment tx
```
marktoda, line 8: looks much better - thanks!

src/coin/xtz/multisigUtils.ts
CapnMigraine: Yes, operations are Michelson code converted to JSON. We only have the path to the parameter.

src/coin/baseCoin/enum.ts
CapnMigraine: We're using `AccountUpdate` now. We can use it for transactions such as public key revelation in Tezos or other coin specific type of transactions, like trustline management transactions in Stellar.

src/coin/xtz/transaction.ts
```diff
@@ -111,7 +112,13 @@ export class Transaction extends BaseTransaction {
    *      originated address
    */
   private async recordOriginationOpFields(operation: OriginationOp, index: number): Promise<void> {
-    this._type = TransactionType.WalletInitialization;
+    const originationData = getOriginationDataFromOperation(operation);
+    if (originationData.contractAddress && Utils.isValidOriginatedAddress(originationData.contractAddress)) {
```
argjv, line 116: Is `Utils.isValidOriginatedAddress(originationData.contractAddress)` necessary? Because if it has the `contractAddress` attribute but it is not a KT1 address, then it is wrong to assume that it is a `WalletInitialization` tx, or not?
We can just remove it and let it be cataloged as an `AddressInitialization` tx.

src/coin/xtz/transactionBuilder.ts
```diff
@@ -266,15 +279,29 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
     this._counter = new BigNumber(counter);
   }
+
+  /**
+   * Set the destination address of a forwarder contract
+   * Used in forwarder contract deployment as destination address
+   *
+   * @param {string} contractAddress - contract address to use
+   */
+  forwarderDestinationAddress(contractAddress: string): void {
```
argjv, line 289: s/forwarderDestinationAddress/forwarderDestination/g
Destination already implies it is an address

src/coin/xtz/transactionBuilder.ts
```diff
@@ -266,15 +279,29 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+    if (!isValidOriginatedAddress(contractAddress)) {
```
argjv, line 290: Before this, check the transaction type to be `AddressInitialization`

test/unit/coin/xtz/transactionBuilder.ts
```diff
@@ -95,6 +95,24 @@ describe('Tezos Transaction builder', function() {
...
+        pub:
+          'xpub661MyMwAqRbcFhCvdhTAfpEEDV58oqDvv65YNHC686NNs4KbH8YZQJWVmrfbve7aAVHzxw8bKFxA7MLeDK6BbLfkE3bqkvHLPgaGHHtYGeY',
+      };
+      const keyPair = new Xtz.KeyPair(source);
+      txBuilder.source(keyPair.getAddress());
+      txBuilder.publicKeyToReveal(keyPair.getExtendedKeys().xpub);
+      txBuilder.counter('0');
+      txBuilder.branch('BM8QdZ92VyaH1s5nwAF9rUXjiPZ3g3Nsn6oYbdKqj2RgHxvWXVS');
+      const tx = await txBuilder.build();
+      tx.type.should.equal(TransactionType.AddressInitialization);
```
argjv, line 114: Check the other attributes including the `forwarderDestiantion`

CapnMigraine: done

src/coin/xtz/multisigUtils.ts
```diff
@@ -965,3 +965,223 @@ const genericMultisig = [
...
+  contractAddress: string,
+  counter: string,
+  source: string,
+  fee: string,
+  gasLimit: string,
+  storageLimit: string,
+  balance: string,
+): OriginationOp {
+  return {
+    kind: 'origination',
```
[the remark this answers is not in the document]
CapnMigraine: done

src/coin/xtz/multisigUtils.ts
```diff
@@ -965,3 +930,286 @@ const genericMultisig = [
...
+ * @param {string} storageLimit Maximum amount in mutez to spend in storage fees
+ * @param {string} balance New multisig account initial balance taken from the source account
+ * @param {object} [operationParams] parameters specific to the type of contract being deployed
+ * @param {string[]} [operationParams.pubKeys] List of public keys of the multisig owner
+ * @param {string} [operationParams.delegate] Optional implicit address to delegate the wallet funds to
+ * @param {number} [operationParams.threshold] Minimum number of signatures required to authorize a multisig operation
+ * @param {string} [operationParams.forwarderDestination] - destination address of new Forwarder contract
+ * @returns {OriginationOp} An origination operation
+ */
+export function originationOperation(
```
argjv, line 1169: Actually, this is more complicated. Let's go back to the 2 methods, mainly so we don't have a set of optional parameters that are incompatible and we have to take care of.

CapnMigraine: done

PR #101, opened 2020-05-15
BGA-10 serialized wallet init new structure

gonzaalvarez70:
This is the implementation of story BGA-10 using the new structure. We are pointing towards BGA-11 since we have new changes specific for this story but we used  BGA-11 branch as a baseline.


src/coin/cgld/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+import { ParseTransactionError } from '../baseCoin/errors';
+//TODO: Remove utils and move this to the types file
+/**
+ *
+ * @param {number} num number to be converted to hex
+ * @returns {string} the hex number
+ */
+function fromNumber(num: number): string {
+  const hex = num.toString(16);
+  return hex.length % 2 === 0 ? '0x' + hex : '0x0' + hex;
```
marktoda, line 13: - s/fromNumber/numberToHexString/g
- you can use BigNumber for operations like this, to keep them simple: `new BigNumber(number).toString(16)`

src/coin/cgld/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ *
+ * @param {string} hex The hex string to be converted
+ * @returns {number} the resulting number
+ */
+function toNumber(hex: string): number {
```
marktoda, line 21: ditto above:
- s/toNumber/hexStringToNumber/g
- new BigNumber(hexSTring, 16).toFixed()

src/coin/cgld/transaction.ts
```diff
@@ -11,4 +12,11 @@ export class Transaction extends Eth.Transaction {
   setTransactionData(txData: TxData): void {
     this._ethTransaction = CgldTransaction.fromJson(txData);
   }
+
+  /**@inheritdoc */
+  public static fromSerialized(coinConfig: Readonly<CoinConfig>, serializedTx: string): Transaction {
+    const tx = new Transaction(coinConfig);
+    tx.setTransactionData(Utils.deserialize(serializedTx)); //TODO: maybe create a constructor that takes 2 arguments
```
marktoda, line 19: +1 constructor with 2 arguments would be good

src/coin/cgld/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  return parseInt(hex.slice(2), 16);
+}
+
+/**
+ *
+ * @param {string} serializedTx the serialized transaction
+ * @returns {TxData} the deserialized transaction
+ */
+export function deserialize(serializedTx: string): TxData {
+  try {
```
marktoda, line 31: silvia mentioned this code came from some library -- can you add a link to it in a comment so we know where to look if it ever breaks?

src/coin/eth/transactionBuilder.ts
```diff
@@ -77,13 +77,36 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
+      case TransactionType.WalletInitialization:
+        this.fee({ fee: transactionJson.gasPrice, gasLimit: transactionJson.gasLimit });
+        this.counter(transactionJson.nonce);
+        this.chainId(Number(transactionJson.chainId));
+        const owners = Utils.decodeWalletCreationData(transactionJson.data);
+        owners.forEach(element => {
+          this.owner(element);
+        });
+        break;
+      //TODO: Add other cases of deserialization
```
marktoda, line 106: Can you add a `default` case that throws? just for a sanity check

src/coin/cgld/transaction.ts
gonzaalvarez70: I solved that in the next commit.

src/coin/cgld/utils.ts
gonzaalvarez70: Sure, I'll add the link.

src/coin/eth/transactionBuilder.ts
gonzaalvarez70: Sure, I will add it to the next commit.

src/coin/cgld/utils.ts
gonzaalvarez70: I'm using just number because the chainId is always an integer. Do you think that we should use BigNumber instead?

gonzaalvarez70: I changed the method's name, it's now located under eth/utils.ts

src/coin/cgld/utils.ts
gonzaalvarez70: That actually returns a string. I'm leaving it as it is for now, do you think that we should use BigNumber there as well? chainId is always an integer. That's even how the celo contract kit does it.

gonzaalvarez70: I changed the method's name, it's now located under eth/utils.ts

test/unit/coin/eth/transactionBuilder.ts
```diff
@@ -105,6 +105,58 @@ describe('Eth Transaction builder', function() {
       should.equal(txJson.nonce, 0);
       should.equal(txJson.chainId, 42);
     });
+
+    it('an unsigned init transaction from serialized', async () => {
```
marktoda, line 109: We should dedupe these test suites as well -- they all test the same functionality, maybe just different input / output data to check against. Can we make a ticket for that @gonzaalvarez70 ?

gonzaalvarez70: Sure, task created as ticket BGA-116

PR #104, opened 2020-05-18
Merge CgldTransaction and CeloTransaction

marktoda:
Since Cgld transactions have slightly different restrictions than ETH
transactions, the EThereumjs-tx library doesn't quite work for
ser/deser/signing of these transactions. This commit merges the existing
fix for this -- CeloTransaction, which implemented the same interface as
Ethereumjs-tx, with the existing CgldTransaction. Since it was not
intended to be exposed outside of a few functions (from/to
JSON/serialized), we could simplify and merge the class into a few
private helper functions


marktoda:
Note this is just a PR to share my idea with @FabianSimtlix

PR #106, opened 2020-05-20
Celo transfer transaction from serialized

silalvado:
This PR includes de-serialization of transfer transaction


src/coin/eth/transactionBuilder.ts
```diff
@@ -108,6 +108,15 @@ export class TransactionBuilder extends BaseTransactionBuilder {
           this.owner(element);
         });
         break;
+      case TransactionType.Send:
+        this.fee({ fee: transactionJson.gasPrice, gasLimit: transactionJson.gasLimit });
+        this.counter(transactionJson.nonce);
+        this.chainId(Number(transactionJson.chainId));
```
marktoda, line 114: you can pull these out to above the `switch` since they will be parsed the same way from all txs

src/coin/eth/transactionBuilder.ts
```diff
@@ -108,6 +108,15 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
         });
         break;
+      case TransactionType.Send:
+        this.fee({ fee: transactionJson.gasPrice, gasLimit: transactionJson.gasLimit });
+        this.counter(transactionJson.nonce);
+        this.chainId(Number(transactionJson.chainId));
+        this._transfer = new TransferBuilder(transactionJson.data);
+        if (transactionJson.to === undefined) {
+          throw new BuildTransactionError('Undefined recipient address');
+        } else this._contractAddress = transactionJson.to;
```
marktoda, line 118: Please always use `{}` for if statements

src/coin/eth/transactionBuilder.ts
```diff
@@ -108,6 +108,15 @@ export class TransactionBuilder extends BaseTransactionBuilder {
           this.owner(element);
         });
         break;
+      case TransactionType.Send:
+        this.fee({ fee: transactionJson.gasPrice, gasLimit: transactionJson.gasLimit });
+        this.counter(transactionJson.nonce);
+        this.chainId(Number(transactionJson.chainId));
+        this._transfer = new TransferBuilder(transactionJson.data);
+        if (transactionJson.to === undefined) {
```
marktoda, line 116: When would this ever be the case? It seems to me that `send` txs should always have a recipient by the nature of `send`. I guess you just have to do this because `to` is `string | undefined` type, right? Not a big deal, but in that case, this is maybe more readable:
```
assert(transactionJson.to); // send transactions should always have recipient defined
this._contractAddress = transactionJson.to!;
```

src/coin/eth/transactionBuilder.ts
```diff
@@ -331,16 +340,21 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     else throw new BuildTransactionError('Invalid address: ' + address);
   }
 
-  transfer(amount: string): TransferBuilder {
+  transfer(amount?: string): TransferBuilder {
```
marktoda, line 343: when would a caller ever want to transfer with undefined amount? I don't see the point for that

src/coin/eth/transactionBuilder.ts
```diff
@@ -331,16 +340,21 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     else throw new BuildTransactionError('Invalid address: ' + address);
   }
```
marktoda, line 342: jsdoc

src/coin/eth/transactionBuilder.ts
```diff
@@ -331,16 +340,21 @@ export class TransactionBuilder extends BaseTransactionBuilder {
...
     if (this._type === TransactionType.Send) {
-      this._transfer = new TransferBuilder().amount(amount);
+      this._transfer = this._transfer || new TransferBuilder();
+      if (amount) {
+        this._transfer.amount(amount);
+      }
       return this._transfer;
     }
     throw new BuildTransactionError('Transfers can only be set for send transactions');
   }
```
marktoda, line 353: jsdoc

src/coin/eth/transactionBuilder.ts
```diff
@@ -331,16 +340,21 @@ export class TransactionBuilder extends BaseTransactionBuilder {
     else throw new BuildTransactionError('Invalid address: ' + address);
   }
 
-  transfer(amount: string): TransferBuilder {
+  transfer(amount?: string): TransferBuilder {
     if (this._type === TransactionType.Send) {
```
marktoda, line 344: nit: personal style preference, but I like to check for invalidities first, and early throw, like
```
if (something is wrong) {
  throw something is wrong
}

if (something else is wrong) {
 throw also
}

everything is good, do stuff
```

src/coin/eth/transferBuilder.ts
```diff
@@ -12,57 +13,66 @@ export class TransferBuilder {
...
   private _signKey: string;
   private _expirationTime: number;
-  private _signature: string;
+  private _signature?: string;
 
   //initialize with default values for non mandatory fields
-  constructor() {
-    this._expirationTime = this.getExpirationTime();
-    this._data = '0x';
+  constructor(serializedData?: string) {
```
marktoda, line 19: Can we keep the no-argument constructor and instead add a `from` builder method, similar to in `transactionbuilder`?

src/coin/eth/transferBuilder.ts
```diff
@@ -104,7 +114,14 @@ export class TransferBuilder {
     return ethUtil.bufferToHex(EthereumAbi.soliditySHA3(...operationData));
   }
 
-  private ethSignMsgHash(): void {
+  private get _encodedSignature(): string {
```
marktoda, line 117: jsdoc

marktoda: Also, what is the point of this method? Seems that it does the same thing as `ethSignMsgHash` and also updates the internal `signature` field. That's a bit odd for a `get` function to be updating state, so I think it'd be better to remove this function and have any callsites that want to update state do that themselves

src/coin/eth/transactionBuilder.ts
FabianSimtlix: done

src/coin/eth/transactionBuilder.ts
FabianSimtlix: rearrange the code to validate to first. Leave the if statement so we can customize the exception message

src/coin/eth/transactionBuilder.ts
FabianSimtlix: remove the else

src/coin/eth/transactionBuilder.ts
FabianSimtlix: This is to allow user to access and update the transfer builder when it already has values. For example after deserialization this builder most likely already has an amount value set

marktoda: Ahh ok that makes sense. In that case, let's remove it from the constructor and add a builder method for it, since it is optional. Builder constructors should only take required paramters

src/coin/eth/transactionBuilder.ts
FabianSimtlix: I prefer to check for validity. So the pattern will be: if (isValid) then do stuff else throw something
I changed it to this style so we all are consistent across the code base.

src/coin/eth/transactionBuilder.ts
FabianSimtlix: Done.
Also change the '||' for an if statement to make it more clear and to be consistent with other classes style

src/coin/eth/transferBuilder.ts
```diff
@@ -12,57 +13,68 @@ export class TransferBuilder {
...
   private _expirationTime: number;
-  private _signature: string;
+  private _signature?: string;
 
-  //initialize with default values for non mandatory fields
-  constructor() {
-    this._expirationTime = this.getExpirationTime();
-    this._data = '0x';
+  constructor();
+  constructor(serializedData: string);
```
marktoda, line 19: What is the point of these? I'm not 100% familiar with this syntax and don't see how it's different from the one optional argument

src/coin/eth/transactionBuilder.ts
marktoda: Still needing a jsdoc here

src/coin/eth/transferBuilder.ts
marktoda: thoughts on this idea?

src/coin/eth/transferBuilder.ts
```diff
@@ -104,7 +116,19 @@ export class TransferBuilder {
...
+  /**
+   * If a signing key is set for this builder, recalculates the signature
+   *
+   * @returns the signature value
+   */
+  private getSignature(): string {
+    if (this._signKey) {
+      this._signature = this.ethSignMsgHash();
+    }
+    return this._signature!; //should it fail if a signature has not being set?
```
marktoda, line 128: I think it should fail if it can't generate a signature. 

It also still seems a little odd to me to have a `get*` function mutating the state of the class. Why not just call it `sign()`?

PR #107, opened 2020-05-20
Fix Wallet Creation Bugs

marktoda:
```
Add signature to json output

This commit adds the signature data to the JSON output of transactions
so that we can maintain the signature between serialization and
deserialization

Ticket: BG-21259
```

```
Add missing fields in Eth Transaction

There are some fields in the base transaction class which were not set
in the Eth Transaction class. Most importantly, the _id field which
allows users of the library to get the transaction id of the
transaction. This commit implements a function in EthTransaction which
updates those fields in Eth Transaction

Ticket: BG-21259
```

- fix lint


src/coin/cgld/utils.ts
```diff
@@ -1,8 +1,9 @@
+import { recoverTransaction } from '@celo/contractkit/lib/utils/signing-utils';
 import { RLP } from 'ethers/utils';
+import { addHexPrefix, ecrecover } from 'ethereumjs-util';
```
CapnMigraine, line 3: these aren't used

src/coin/cgld/utils.ts
```diff
@@ -16,9 +17,11 @@ export function deserialize(serializedTx: string): TxData {
   try {
     const rawValues = RLP.decode(serializedTx);
     let chainId = rawValues[9];
```
CapnMigraine, line 19: can we either use an array destructuring assignment or match the keys of the `rawValues` array against an enum so its values can be accessed by property name?

src/coin/eth/utils.ts
```diff
@@ -188,3 +197,43 @@ export function calculateForwarderAddress(contractAddress: string, contractCount
   const forwarderAddress = generateAddress(contractAddress, addHexPrefix(contractCounter.toString(16)));
   return addHexPrefix(forwarderAddress.toString('hex'));
 }
+
+/**
+ * Convert the given signature parts to a string representation
+ *
+ * @param sig The signature to convert to string
+ */
```
CapnMigraine, line 205: nit: set type of `sig` param, add `@returns`

src/coin/eth/utils.ts
```diff
@@ -188,3 +197,43 @@ export function calculateForwarderAddress(contractAddress: string, contractCount
...
+export function toStringSig(sig: SignatureParts): string {
+  return bufferToHex(Buffer.concat([setLengthLeft(sig.r, 32), setLengthLeft(sig.s, 32), toBuffer(sig.v)]));
+}
+
+/**
+ * Convert the given signature parts to a string representation
+ *
+ * @param sig The signature to convert to string
+ */
+export function fromStringSig(sig: string): SignatureParts {
```
CapnMigraine, line 215: nit: same as above. 

This function is not being used, is it needed?

src/coin/eth/utils.ts
```diff
@@ -188,3 +197,43 @@ export function calculateForwarderAddress(contractAddress: string, contractCount
...
+    r: bufferToHex(r),
+    s: bufferToHex(s),
+  };
+}
+
+/**
+ * Return whether or not the given tx data has a signature
+ *
+ * @param txData The transaction data to check for signature
+ * @returns true if the tx has a signature, else false
```
CapnMigraine, line 228: nit: add types

test/unit/coin/eth/utils.ts
```diff
@@ -26,7 +26,7 @@ describe('ETH util library', function() {
     should.equal(getContractData(CONTRACT_ADDRESSES), testData.EXPECTED_CONTRACT_DATA);
   });
 
-  describe('Should validate valid createForwareder Id', function() {
+  describe('Should validate valid createForwarer Id', function() {
```
CapnMigraine, line 29: nit: typo :)

```suggestion
  describe('Should validate valid createForwarder Id', function() {
```

src/coin/cgld/utils.ts
marktoda: good idea

src/coin/eth/utils.ts
marktoda: will remove -- i thought I'd need it but guess not :)

src/coin/eth/utils.ts
marktoda: done but is this really necessary in typescript?

test/unit/coin/eth/utils.ts
marktoda: LOL

src/coin/eth/utils.ts
CapnMigraine: No, I've been adding them for consistency with pre-typescript functions. I hadn't found any recommendations against it [until now](https://dev.to/kamranayub/why-you-don-t-need-types-in-jsdoc-when-documenting-typescript-1pb0).

You can remove them

PR #111, opened 2020-05-22
Improve transaction builder speed

marktoda:
The ETH transaction builder was very slow for wallet initialization
transactions because we were parsing data (which can be huge) using
BigNumber, which is slower the bigger the number gets. In fact, we could
keep the data parsing logic very simple here as we just need the hex
string.

Ticket: BG-21437


marktoda:
Before: test suite took 2 minutes and 30 seconds: https://cloud.drone.io/BitGo/bitgo-account-lib/858/2/5
After: test suite took 1 minute and 8 seconds: https://cloud.drone.io/BitGo/bitgo-account-lib/862/2/5

PR #113, opened 2020-05-22
BGA-13 Send token

gonzaalvarez70:
Initial implementation for sendMultisigToken transaction in account-lib for Celo.


src/coin/eth/index.ts
```diff
@@ -2,6 +2,8 @@ export { KeyPair } from './keyPair';
 export { Transaction } from './transaction';
 export { TransactionBuilder } from './transactionBuilder';
 export { TransferBuilder } from './transferBuilder';
```
marktoda, line 4: Do we still need to export `TransferBuilder` now that it is sort-of abstract?

src/coin/eth/transactionBuilder.ts
```diff
@@ -128,6 +132,13 @@ export class TransactionBuilder extends BaseTransactionBuilder {
           this.owner(element);
         });
         break;
+      case TransactionType.Send:
+        if (transactionJson.to === undefined) {
+          throw new BuildTransactionError('Undefined recipient address');
+        }
+        this._contractAddress = transactionJson.to;
+        this._transfer = new TransferFundsBuilder(transactionJson.data);
+        break;
```
marktoda, line 141: Should you add a case here for `SendToken` as well?

src/coin/eth/transferFundsBuilder.ts
```diff
@@ -0,0 +1,86 @@
+import ethUtil from 'ethereumjs-util';
+import EthereumAbi from 'ethereumjs-abi';
+import BigNumber from 'bignumber.js';
+import { BuildTransactionError } from '../baseCoin/errors';
+import { sendMultiSigData } from './utils';
+import { sendMultisigMethodId } from './walletUtil';
+import { TransferBuilder } from './';
+
+/** ETH transfer builder */
+export class TransferFundsBuilder extends TransferBuilder {
```
marktoda, line 10: Instead of having two entirely different builder classes for transferring native assets and tokens, what if we just kept the one `TransferBuilder` and added a `coin` field (and builder function) that:
- defaults to the native currency (cgld, rbtc, etc)
- if the user inputs anything other than the native currency, it checks it against @bitgo/statics coin list: https://github.com/BitGo/BitGoJS/tree/master/modules/statics
- using the @bitgo/statics coin details, it can determine if the given asset is a token -- if it's a token then we can make the necessary changes to the build flow: use the token `operation hash` and the token function call.

What do you think? I think that's a simpler user API and also probably simpler code. cc @CapnMigraine what are you guys doing for XTZ in this situation?

src/coin/eth/utils.ts
```diff
@@ -98,6 +90,32 @@ export function sendMultiSigData(
...
+export function sendMultiSigTokenData(
+  to: string,
+  value: number,
+  tokenContractAddress: string,
+  expireTime: number,
+  sequenceId: number,
+  signature: string,
+): string {
+  const params = [to, value, tokenContractAddress, expireTime, sequenceId, toBuffer(signature)];
+  const types = ['address', 'uint', 'address', 'uint', 'uint', 'bytes'];
```
marktoda, line 113: can you save the types in `walletUtil` like we have done for the constructor args

src/coin/eth/index.ts
gonzaalvarez70: You are right we don't need it anymore.

src/coin/eth/transactionBuilder.ts
gonzaalvarez70: That's part of story BGA-14

src/coin/eth/utils.ts
gonzaalvarez70: Sure, sounds good.

src/coin/eth/transferFundsBuilder.ts
gonzaalvarez70: With that approach, I'm worried about the token contract address, also for deserialization I think that keeping those transactions (sendMultiSig and sendMultiSigToken) separately would be better. I think that we could think that approach as an extension once we start implementing RSK. What do you think?

FabianSimtlix: I also think that it's better if we separate the builders. That way we can keep each one focus on a particular contract method. The logic that you mentioned would have more sense in transactionBuilder as an strategy there to choose which builder to use based on the given asset.
Also it relates with next changes where we want to split transactions by its type and payload

marktoda: > I'm worried about the token contract address, also for deserialization

That's the point of `@bitgo/statics` -- it contains coin name to contract address mappings, which you can use both in ser and deser. 

> That way we can keep each one focus on a particular contract method

The difference between each contract method is quite small, and shouldn't affect any part of the builder other than the serialization of the tx

This idea is also what XTZ are doing: https://github.com/BitGo/bitgo-account-lib/blob/e479643c68251d2d8e25235cc95f6c19eebdb646/src/coin/xtz/transferBuilder.ts#L23

marktoda: Here is how you can get the contract address from @bitgo/statics:

```
import { coins, Erc20Coin } from '@bitgo/statics';

const omg = coins.get('omg');
if (omg instanceof Erc20Coin) {
  console.log(omg.contractAddress);
}
```

gonzaalvarez70: What do you think about keeping both TransferBuilders but in TransferTokenBuilder instead of manually setting the tokenContractAddress we create a single transfer method in transactionBuilder that receives a "coin" parameter set as the transaction coin type by default (cgld, rsk, eth).
Based on that parameter we infer the proper transfer builder using what we got in that parameter, and if needed the tokenContractAddress from @bitgo/statics. This way we can keep the structure that makes both transfer builder more maintainable.

However I think that cUSD (the main Celo network token) wasn't added to statics yet, so we would need to make that change first to properly infer the contractAddress.

marktoda: Feel free to submit PRs into @bitgo/statics -- it is meant to be the canonical source of "coins that BitGo supports".

That idea sounds fine -- I still think there is no need for 2 separate builders which are doing wholly the same thing, but, it's OK

gonzaalvarez70: The idea of two builders is to handle the different fields that each contract call has (sendMultiSig has a data field and sendMultiSigToken has a tokenContractAddress).
And this way those are more maintainable in case we need to add new functionality to handle those transactions since they are different transaction types. Keeping both in one file would cause a few methods to make two different things because the implementation for signing or decoding is different.
Does that make sense to you?

marktoda: No, I still disagree, but it's not worth the arguemnt -- just continue on

gonzaalvarez70: No problem, we can just continue with the initial comment idea (just  one `TransferBuilder ` and added a `coin` field), we just thought that keeping both transferBuilder could be valid even when using the new coin field.

src/coin/eth/transferBuilder.ts
```diff
@@ -68,14 +88,25 @@ export class TransferBuilder {
...
-        new BigNumber(this._amount).toNumber(),
-        this._data,
-        this._expirationTime,
-        this._sequenceId,
-        this.getSignature(),
-      );
+      if (this._tokenContractAddress !== undefined) {
+        return sendMultiSigTokenData(
+          this._toAddress,
+          new BigNumber(this._amount).toNumber(),
```
marktoda, line 94: This is lossy and throws when given a big number. Please keep the amount as either a string or bignumber

src/coin/eth/transferBuilder.ts
```diff
@@ -68,14 +88,25 @@ export class TransferBuilder {
...
+          this._tokenContractAddress,
+          this._expirationTime,
+          this._sequenceId,
+          this.getSignature(),
+        );
+      } else {
+        return sendMultiSigData(
+          this._toAddress,
+          new BigNumber(this._amount).toNumber(),
```
marktoda, line 103: ditto above

src/coin/eth/transferBuilder.ts
```diff
@@ -91,6 +122,41 @@ export class TransferBuilder {
...
   }
 
+  protected getOperationHash(): (string | Buffer)[][] {
+    const operationData = this.getOperationData();
+    return ethUtil.bufferToHex(EthereumAbi.soliditySHA3(...operationData));
+  }
+
+  private getOperationData(): any[][] {
+    let operationData;
+    if (this._tokenContractAddress !== undefined) {
```
marktoda, line 132: Readability Nit: You can save the `@bitgo/statics` coin object instead of just the token address, so that you can just check `if (this.coin.isToken)` or `if (this.coin instanceof Erc20Coin)` here, just for readability.. up to you. This may also be useful for CELO where even the native coins have token addresses

src/coin/baseCoin/enum.ts
```diff
@@ -3,6 +3,7 @@
  */
 export enum TransactionType {
   Send,
+  SendToken,
```
gonzaalvarez70, line 6: In my opinion, the new transaction type is not needed anymore. What do you think @marktoda  should we remove it?

src/coin/eth/transferBuilder.ts
```diff
@@ -91,6 +122,41 @@ export class TransferBuilder {
     );
   }
 
+  protected getOperationHash(): (string | Buffer)[][] {
+    const operationData = this.getOperationData();
+    return ethUtil.bufferToHex(EthereumAbi.soliditySHA3(...operationData));
+  }
+
+  private getOperationData(): any[][] {
```
marktoda, line 130: Can you define the return type as something better than `any`? Does `(string | Buffer)[][]` not work anymore? I'd also consider thinking about merging this and `getOperationHash` as this operation data will always be just going straight into `soliditySHA3`

src/coin/eth/transferBuilder.ts
```diff
@@ -91,6 +122,41 @@ export class TransferBuilder {
     );
   }
 
+  protected getOperationHash(): (string | Buffer)[][] {
```
marktoda, line 125: wait how is this return type correct? Doesn't it just return a hex string?

src/coin/eth/utils.ts
```diff
@@ -92,9 +86,33 @@ export function sendMultiSigData(
...
+ * @param {number} value Amount to tranfer
+ * @param {string} tokenContractAddress the address of the erc20 token contract
+ * @param {number} expireTime expiration time for the transaction in seconds
+ * @param {number} sequenceId sequence id
+ * @param {string} signature signature of the call
+ * @returns {string} -- the contract method encoded data
+ */
+export function sendMultiSigTokenData(
+  to: string,
+  value: number,
```
marktoda, line 107: lossy - string or bignumber for values please

test/unit/coin/eth/transferTokenBuilder.ts
```diff
@@ -0,0 +1,55 @@
...
+        .coin('tcusd')
+        .expirationTime(1590078260)
+        .amount(amount)
+        .to(toAddress)
+        .contractSequenceId(2)
+        .key(key);
+      const result = builder.signAndBuild();
+      should.equal(result, testData.SEND_TOKEN_DATA);
+    });
+    it('should fail if a coin is not an ERC20 instance', () => {
```
marktoda, line 21: nit spaces between these it clauses

test/unit/coin/eth/transferTokenBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import should from 'should';
+import { TransferBuilder } from '../../../../src/coin/eth';
+import * as testData from '../../../resources/eth/eth';
+
+describe('Eth send token multi sig builder', function() {
+  describe('should build', () => {
+    const toAddress = '0x7325A3F7d4f9E86AE62Cf742426078C3755730d5';
+    const key = '8CAA00AE63638B0542A304823D66D96FF317A576F692663DB2F85E60FAB2590C';
+    const amount = '0.01';
```
marktoda, line 9: wait, we define amount as big units (i.e. with decimal places?) I expected that we would be using base units (i.e. never decimal places). Where are you converting to base units as I don't see that anywhere above?

marktoda: Note generally our external APIs accept base units, just to keep things simple. This _could_ be OK, but I have a hunch that the value is not being converted properly

test/unit/coin/eth/transferTokenBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import should from 'should';
+import { TransferBuilder } from '../../../../src/coin/eth';
+import * as testData from '../../../resources/eth/eth';
+
+describe('Eth send token multi sig builder', function() {
```
marktoda, line 5: now that theres only 1 builder, can we just add this to another describe block in `test/transferBuilder`?

test/unit/coin/eth/transferTokenBuilder.ts
```diff
@@ -0,0 +1,55 @@
+import should from 'should';
+import { TransferBuilder } from '../../../../src/coin/eth';
+import * as testData from '../../../resources/eth/eth';
+
+describe('Eth send token multi sig builder', function() {
+  describe('should build', () => {
+    const toAddress = '0x7325A3F7d4f9E86AE62Cf742426078C3755730d5';
+    const key = '8CAA00AE63638B0542A304823D66D96FF317A576F692663DB2F85E60FAB2590C';
```
marktoda, line 8: Please also add tests that test using `xpubs` (as that's what is used in bitgo more often than raw keys)

test/unit/coin/eth/transferTokenBuilder.ts
```diff
@@ -0,0 +1,55 @@
...
+        .contractSequenceId(2)
+        .key(key);
+      const result = builder.signAndBuild();
+      should.equal(result, testData.SEND_TOKEN_DATA);
+    });
+    it('should fail if a coin is not an ERC20 instance', () => {
+      should.throws(() => {
+        new TransferBuilder().coin('eth');
+      }, 'There was an error using that coin as a transfer currency');
+    });
```
marktoda, line 25: We can think of some other edge cases to test here, hopefully to avoid finding issues later on. Some ideas:
- a coin that isn't in @bitgo/statics at all
- a coin that is a token in @bitgo/statics but not an erc20 token (https://github.com/BitGo/BitGoJS/blob/11b60b28046f8800f67d02fe235b85c15e449f5d/modules/statics/src/coins.ts#L354)
- invalid key, sequence id, destination, amount (not just missing), expiration
- success with very large amount, sequence id

src/coin/baseCoin/enum.ts
marktoda: :+1:

src/coin/eth/transferBuilder.ts
gonzaalvarez70: Sure, I'll go with this option could be useful in the future to have that coin stored.

src/coin/eth/transferBuilder.ts
gonzaalvarez70: Yeah I will change that.

src/coin/eth/transferBuilder.ts
gonzaalvarez70: You are right, this actually returns a string.

test/unit/coin/eth/transferTokenBuilder.ts
gonzaalvarez70: Sure

test/unit/coin/eth/transferTokenBuilder.ts
gonzaalvarez70: I just changed this, we are not taking numbers with decimal places with the latest fix for BGA-120 as I understand.

test/unit/coin/eth/transferTokenBuilder.ts
gonzaalvarez70: You're right. Done.

test/unit/coin/eth/transferTokenBuilder.ts
gonzaalvarez70: Ok I'll add that.

gonzaalvarez70: Can we derive a private key from an xpub? Maybe we need to use an xprv. Am I right?

test/unit/coin/eth/transferTokenBuilder.ts
gonzaalvarez70: Ok I'll add some new test cases sounds good.

src/coin/eth/transferBuilder.ts
gonzaalvarez70: Solved as part of BGA-120

src/coin/eth/transferBuilder.ts
gonzaalvarez70: Solved as part of BGA-120

src/coin/eth/transferBuilder.ts
marktoda: I'm just bewildered that it compiled like that before, hmm

test/unit/coin/eth/transferTokenBuilder.ts
marktoda: ah yep xprv

gonzaalvarez70:
> LGTM, though tests seem broken

We would actually need for https://github.com/BitGo/BitGoJS/pull/753 to be merged and add that new @bitgo/statics reference in order to make the tests work.

src/coin/eth/transferBuilder.ts
```diff
@@ -94,6 +137,47 @@ export class TransferBuilder {
...
+  private getOperationData(): (string | Buffer)[][] {
+    let operationData;
+    if (this._tokenContractAddress !== undefined) {
+      operationData = [
+        ['string', 'address', 'uint', 'address', 'uint', 'uint'],
+        [
+          'ERC20',
+          new ethUtil.BN(ethUtil.stripHexPrefix(this._toAddress), 16),
+          this._amount,
+          new ethUtil.BN(ethUtil.stripHexPrefix(this._tokenContractAddress), 16),
```
twtaylorbitgo, line 161: if the only variation for this is one parameter, you should be writing an inline ternary statement to encapsulate this or just varying that one parameter. this is a lot of extra lines of code here

src/coin/eth/transferBuilder.ts
```diff
@@ -148,20 +217,38 @@ export class TransferBuilder {
...
+    if (!(data.startsWith(sendMultisigMethodId) || data.startsWith(sendMultisigTokenMethodId))) {
       throw new BuildTransactionError(`Invalid transfer bytecode: ${data}`);
     }
-    const splitBytecode = data.split(sendMultisigMethodId);
-    if (splitBytecode.length !== 2) {
-      throw new BuildTransactionError(`Invalid send bytecode: ${data}`);
+    let decoded;
+    if (this.isTokenTransfer(data)) {
+      decoded = this.getRawDecoded(sendMultiSigTokenTypes, this.getBufferedByteCode(sendMultisigTokenMethodId, data));
+      this._tokenContractAddress = ethUtil.bufferToHex(decoded[2]);
```
twtaylorbitgo, line 227: instead of a 2, can you put this in a constant index describing what this field describes? 

https://stackoverflow.com/questions/4117214/using-constants-as-indices-for-javascript-associative-arrays

gonzaalvarez70: Sounds good, I'll create a new Enum to describe this index.

src/coin/eth/transferBuilder.ts
gonzaalvarez70: The first array of field types is different and the second one has two different fields. I could do it in a more compact way but I thought that this way would help readability.

gonzaalvarez70: The only way that I can think about is the following. Let me know if it sounds good
`  private getOperationData(): (string | Buffer)[][] {
    const operationData = [
      this.isTokenOperation()
        ? ['string', 'address', 'uint', 'address', 'uint', 'uint']
        : ['string', 'address', 'uint', 'bytes', 'uint', 'uint'],
      [
        this.isTokenOperation() ? 'ERC20' : 'ETHER',
        new ethUtil.BN(ethUtil.stripHexPrefix(this._toAddress), 16),
        this._amount,
        this.isTokenOperation()
          ? new ethUtil.BN(ethUtil.stripHexPrefix(this._tokenContractAddress), 16)
          : new Buffer(ethUtil.stripHexPrefix(this._data) || '', 'hex'),
        this._expirationTime,
        this._sequenceId,
      ],
    ];
    return operationData;
  }

  private isTokenOperation(): boolean {
    return this._tokenContractAddress !== undefined;
  }`

src/coin/eth/transferBuilder.ts
twtaylorbitgo: thanks

src/coin/eth/transferBuilder.ts
twtaylorbitgo: ahhh I see. the bytes variation. Ok yeah I think your original is way more readable. Gonna dismiss this, thanks

PR #116, opened 2020-05-27
BGA-120 Fix bug for large amount

ipelaez-simtlix:
Ethereumjs-abi decode can internally convert string to long number


src/coin/eth/transferBuilder.ts
```diff
@@ -155,7 +158,7 @@ export class TransferBuilder {
     const serializedArgs = Buffer.from(splitBytecode[1], 'hex');
     const decoded = EthereumAbi.rawDecode(['address', 'uint', 'bytes', 'uint', 'uint', 'bytes'], serializedArgs);
     this._toAddress = ethUtil.bufferToHex(decoded[0]);
-    this._amount = ethUtil.bufferToInt(decoded[1]).toString();
+    this._amount = new BigNumber(ethUtil.bufferToHex(decoded[1])).toString();
```
marktoda, line 161: Please use `toFixed()` instead of `toString()` because a large number will use exponential notation with `toString()`:

> To get the string value of a BigNumber use toString() or toFixed(). Using toFixed() prevents exponential notation being returned, no matter how large or small the value.


```
> new BigNumber('1234000000000000000000').toString()
'1.234e+21'
> new BigNumber('1234000000000000000000').toFixed()
'1234000000000000000000'

```

PR #120, opened 2020-06-01
BGA-129 forwarder transaction

gonzaalvarez70:


src/coin/eth/types.ts
```diff
@@ -8,7 +8,7 @@ import { KeyPair } from './keyPair';
  * An Ethereum transaction with helpers for serialization and deserialization.
  */
 export class EthTransactionData implements EthLikeTransactionData {
-  constructor(private tx: EthereumTx, protected chainId?: string) {}
+  constructor(private tx: EthereumTx, protected deployedAddress?: string, protected chainId?: string) {}
```
marktoda, line 11: Think about changing this signature to include just 2 fields, where the second one includes the various optional parameters. 

Generally it's best to avoid having more than one optional positional parameter like this, as if someone wanted to provide `chainId` but not `deployedAddress` they would have to call it like `new EthTransactionData(tx, undefined, chainId)` which is hard to read.

marktoda: If you do this, may as well do it for `CgldTransaction` as well

gonzaalvarez70: Ok, that sounds good, so it would be like an object? or an array of those optional params?

marktoda: in ts I usually define an interface like:

```
interface OptionalTxDataOptions {
   deployedAddress?: string;
  chainId?: string;
}
```

Then your constructor becomes
```
  constructor(private tx: EthereumTx, options: OptionalTxDataOptions) {}

```

gonzaalvarez70: Something like   `constructor(private tx: EthereumTx, protected args?: { deployedAddress?: string; chainId?: string }) {}`

marktoda: thats OK too

src/coin/eth/iface.ts
```diff
@@ -16,6 +16,7 @@ export interface TxData {
   data: string;
   value: string;
   chainId?: string;
+  deployedAddress?: string;
```
twtaylorbitgo, line 19: is there a reason a bunch of these values are nullable? I would think every TX had an id for Eth (ID is really just a hash from the transaction data)

gonzaalvarez70: Each nullable value is optional for a Transaction, if I remember correctly there's a flow where we can create a transaction without ID and calculate it afterward. Since some of the internal values can change. @marktoda  could you please confirm, I think that it's related to how account-lib is used elsewhere.

marktoda: Right, we built transactions without the signature -- you *can* calculate the txid with such an unsigned tx, but it's not very useful and is actually disingenuous because it will change once the tx is signed

twtaylorbitgo: ah ok - yeah I'll connect with him then. thanks.

PR #122, opened 2020-06-02
Bga 116 test suite enhancement

silalvado:
Enhancement of unit tests files structure.


test/unit/coin/eth/transactionBuilder/addressInitialization.ts
```diff
@@ -0,0 +1,58 @@
...
+  });
+
+  it('should build properly and return a correct address', async () => {
+    const txBuilder = getBuilder('eth') as Eth.TransactionBuilder;
+    txBuilder.type(TransactionType.AddressInitialization);
+    txBuilder.fee({
+      fee: '10',
+      gasLimit: '1000',
+    });
+    txBuilder.chainId(31);
```
twtaylorbitgo, line 48: what is 31? I don't see it here: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md

marktoda: https://github.com/ethereum-lists/chains/blob/master/_data/chains/31.json

marktoda:
I thought that we were planning on deduplicating the tests across etc/cgld/rbtc/eth in general -- they all adhere to the same API, so we should be able to test them all in one file, just changing the classes we are testing, right? I'd suggest something like:

-- test/eth/transactionBuilder.ts --
```
['cgld', 'etc', 'rbtc', 'eth'].forEach(coinName => {
   const txBuilder = getBuilder(coinName);

  ... generic tests against tx builder
}
```

marktoda:
It's OK if that's a future change -- I just don't quite even understand the change going on here :) -- can you add more details at least?

silalvado:
Yes, we will make that change in the next sprint. Now we just change a little the structure of the files by separating the tests from transactionBuilder into three different files to improve their reading, depending on whether it is a send transaction, a wallet initialization transaction or an address initialization transaction.
In addition, we added the use of extended keys.

silalvado:
Merge BGA-116 Test suite enhancement

PR #124, opened 2020-06-04
Bga 122 staking lock

FabianSimtlix:
This PR add functionality to lock gold for celo coin as part of staking process


src/coin/baseCoin/enum.ts
```diff
@@ -18,3 +20,7 @@ export enum AddressFormat {
   hex = 'hex',
   base58 = 'base58',
 }
+
+export enum StakingOperationsTypes {
```
marktoda, line 24: `StakingOperationTypes`

src/coin/cgld/stakingBuilder.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  constructor() {
+    this._type = StakingOperationsTypes.LOCK;
+    this._coinName = mainnet;
+  }
+
+  coin(name: string): this {
+    const coin = coins.get(name);
+    const coinName = coin.name;
+    if (!(coinName === mainnet || coinName === alfajores)) {
```
marktoda, line 21: Use `coin.network` -- it has a `type` field which is either `mainnet` or `testnet` -- source: https://github.com/BitGo/BitGoJS/blob/43ece7ce41b2433adc04b18113aec26879320254/modules/statics/src/networks.ts#L3

src/coin/cgld/stakingBuilder.ts
```diff
@@ -0,0 +1,54 @@
...
+import { alfajores, getOperationParams, mainnet } from './stakingUtils';
+
+export class StakingBuilder {
+  private _amount: string;
+  private _type: StakingOperationsTypes;
+  private _coinName: string;
+
+  constructor() {
+    this._type = StakingOperationsTypes.LOCK;
+    this._coinName = mainnet;
```
marktoda, line 15: It doesn't seem like either of these should be defaults -- just leave them undefined and require that they be defined before building

src/coin/cgld/stakingBuilder.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  coin(name: string): this {
```
marktoda, line 18: Doesn't `TransactionBuilder` already have the coin from its construction? The user shouldn't have to set it twice -- maybe `TransactionBuilder` can just pass it quietly in the constructor of this builder

src/coin/cgld/staking.ts
```diff
@@ -0,0 +1,17 @@
+import EthereumAbi from 'ethereumjs-abi';
+import { addHexPrefix, toBuffer } from 'ethereumjs-util';
+
+export class Staking {
```
marktoda, line 4: I get the idea, but if the only use is the 2 liner `serialize` - this class seems unnecessary

marktoda: Well - it could be useful in the future, but it's definitely not `staking` specific. It seems to be intended for calling any contract _through_ the `WalletSimple` Contract. That could be useful for plenty of other things, not just CGLD and not just staking. If you want to keep it, you can move it to `eth/` folder and change its name to `ContractCall` or `ContractInteraction` or something as that's what it's really representing

src/coin/cgld/stakingBuilder.ts
FabianSimtlix: Yes, we use the _coinConfig field on TransactionBuilder to set the coin when this builder is created. But also leave this method to allow the user override that "default".
This way it's not mandatory to set the coin, but can be done if the user needs to.

src/coin/cgld/stakingUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+export const alfajores = 'tcgld';
+export const mainnet = 'cgld';
+
+export const LockMethodId = '0xf83d08ba';
+
+const operations = {
+  [StakingOperationsTypes.LOCK]: {
+    [alfajores]: {
+      contractAddress: '0x94c3e6675015d8479b648657e7ddfcd938489d0d',
+      methodId: '0xf83d08ba',
```
marktoda, line 12: The methodId and types are the same for the different networks in all cases -- why define them here?

src/coin/cgld/staking.ts
```diff
@@ -0,0 +1,17 @@
...
+  constructor(
+    public amount: string,
+    public address: string,
+    private _methodId: string,
+    private _types: string[],
+    private _params: [],
```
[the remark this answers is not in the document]
FabianSimtlix: Yes, it's an any[] as each method receives different parameters types

src/coin/cgld/transactionBuilder.ts
```diff
@@ -21,4 +36,43 @@ export class TransactionBuilder extends Eth.TransactionBuilder {
     }
     return tx;
   }
+
+  protected setTransactionTypeFields(decodedType: TransactionType, transactionJson: TxData): void {
+    if (decodedType === TransactionType.StakingLock) {
```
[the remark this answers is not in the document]
FabianSimtlix: This is changed on next PR #126

src/coin/cgld/transactionBuilder.ts
```diff
@@ -21,4 +36,43 @@ export class TransactionBuilder extends Eth.TransactionBuilder {
...
+    }
+  }
+
+  //region Stake methods
+  lock(): StakingBuilder {
+    if (this._type !== TransactionType.StakingLock) {
+      throw new BuildTransactionError('Lock can only be set for Staking Lock transactions type');
+    }
+    if (!this._stakingBuilder) {
+      this._stakingBuilder = new StakingBuilder().type(StakingOperationsTypes.LOCK).coin(this._coinConfig.name);
```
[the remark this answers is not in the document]
twtaylorbitgo: this is fine - we should talk about this as a design thing, not a PR thing

these constructors should take interfaces as factories for any default factory construction for testability purposes

PR #125, opened 2020-06-04
Transfer data needs to be built without signature

FabianSimtlix:
remove signature as mandatory value to build transfer data


src/coin/eth/transferBuilder.ts
```diff
@@ -35,6 +35,7 @@ export class TransferBuilder {
       //initialize with default values for non mandatory fields
       this._expirationTime = this.getExpirationTime();
       this._data = '0x';
+      this._signature = '0x';
```
gonzaalvarez70, line 38: Maybe we could create a constant for `EMPTY_HEX` so we use the same for data and the various signature set.

test/resources/eth/eth.ts
```diff
@@ -58,6 +58,9 @@ export const UNSIGNED_TX =
 export const SEND_FUNDS_DATA =
   '0x391252150000000000000000000000007325a3f7d4f9e86ae62cf742426078c3755730d5000000000000000000000000000000000000000000000000016345785d8a000000000000000000000000000000000000000000000000000000000000000000c0000000000000000000000000000000000000000000000000000000005ec6ab34000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000414ef50512a771848411eea2bd0250354e1162b137026b11f6e7267e8d0abda0c87b1cf154db999496c789474552a26869d4fc7dc9f7a04e193b2cfdfa9b608f421b00000000000000000000000000000000000000000000000000000000000000';
 
+export const SEND_FOUNDS_NO_KEY_DATA =
```
gonzaalvarez70, line 61: minor typo: FUNDS instead of FOUNDS

src/coin/eth/transferBuilder.ts
FabianSimtlix: done

src/coin/eth/transferBuilder.ts
```diff
@@ -16,12 +16,13 @@ import { TransferFieldsIndex } from './enum';
 
 /** ETH transfer builder */
 export class TransferBuilder {
+  private readonly _EMPTY_HEX_VALUE = '0x';
```
marktoda, line 19: nit: This is more of a class constant than an instance variable, so can use `static` keyword

PR #126, opened 2020-06-04
BGA-124 Vote staking

gonzaalvarez70:
This PR adds functionality to vote for a validator group on the Celo network as part of the staking process.


src/coin/cgld/stakingUtils.ts
```diff
@@ -18,6 +19,18 @@ const operations = {
       types: [],
     },
   },
+  [StakingOperationsTypes.VOTE]: {
+    [alfajores]: {
+      contractAddress: '0x11fe523f93cac185d12cb39cc3bd279D2de524f8',
+      methodId: '0x580d747a',
+      types: ['address', 'uint256', 'address', 'address'],
```
marktoda, line 26: The methodId and types are the same for the different networks in all cases -- why define them here?

src/coin/cgld/stakingBuilder.ts
```diff
@@ -51,4 +91,34 @@ export class StakingBuilder {
...
+  }
+
+  private decodeStakingData(data: string): void {
+    if (!data.startsWith(VoteMethodId)) {
+      throw new BuildTransactionError(`Invalid staking bytecode: ${data}`);
+    }
+
+    this._type = StakingOperationsTypes.VOTE;
+    const operation = getOperationParams(this._type, this._coinName);
+    const decoded = getRawDecoded(operation.types, getBufferedByteCode(operation.methodId, data));
```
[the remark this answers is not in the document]
twtaylorbitgo: note: this is a bit unsafe and should not be done inline from a function that could return an array with less than 4 items possibly, at the very least we should be checking for that length before destructuring

src/coin/cgld/transactionBuilder.ts
```diff
@@ -38,13 +41,19 @@ export class TransactionBuilder extends Eth.TransactionBuilder {
...
+        this._stakingBuilder = new StakingBuilder()
+          .type(StakingOperationsTypes.LOCK)
+          .amount(transactionJson.value)
+          .coin(this._coinConfig.name);
+        break;
+      case TransactionType.StakingVote:
+        this._stakingBuilder = new StakingBuilder(transactionJson.data);
+        break;
+      default:
+        super.setTransactionTypeFields(decodedType, transactionJson);
```
twtaylorbitgo, line 55: NIT: it's inconsistent how we call the super here but in `getTransactionData()` we just let it drop through the switch - let's just try to make this a little more consistent in terms of style maybe?

src/coin/eth/utils.ts
```diff
@@ -269,3 +271,29 @@ export function hasSignature(txData: TxData): boolean {
...
+
+/**
+ * Get the raw data decoded for some types
+ *
+ * @param {string[]} types ABI types definition
+ * @param {Buffer} serializedArgs encoded args
+ * @returns {Buffer[]} the decoded raw
+ */
+export function getRawDecoded(types: string[], serializedArgs: Buffer): Buffer[] {
+  return EthereumAbi.rawDecode(types, serializedArgs);
```
twtaylorbitgo, line 283: QUESTION/NIT: any idea what this throws if it fails to decode? as in what kind of error will a caller get upstream? it seems like we want to be able to surface meaningful errors

src/coin/cgld/stakingBuilder.ts
```diff
@@ -38,10 +50,38 @@ export class StakingBuilder {
...
+  }
+
   build(): Staking {
     switch (this._type) {
       case StakingOperationsTypes.LOCK:
         return this.buildLockStaking();
+      case StakingOperationsTypes.VOTE:
+        this.validateVoteFields();
+        const params = [this._groupToVote, this._amount, this._lesser, this._greater];
+        return this.buildVoteStaking(params);
```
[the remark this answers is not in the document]
gonzaalvarez70: you are right, that should be moved to buildVoteStaking.

src/coin/cgld/stakingUtils.ts
gonzaalvarez70: Right, it's declared as a constant in the same file. I could just use that value.

src/coin/cgld/stakingBuilder.ts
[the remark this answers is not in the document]
gonzaalvarez70: That sounds good!

src/coin/cgld/transactionBuilder.ts
gonzaalvarez70: This is because the staking transactions are exclusive for Celo, but the rest can be resolved using the general Eth behaviour. What do you suggest? something like checking the different types and then calling the super? That would be checking for the type twice.

src/coin/eth/utils.ts
gonzaalvarez70: it has a few specified errors like:
`throw new Error('Decoded uint exceeds width: ' + parsedType.size + ' vs ' + num.bitLength())`
`  throw new Error('Decimals not supported yet')`
`  throw new Error('Unsupported or invalid type: ' + parsedType.name)`
`
      throw new Error('Decoded int exceeds width: ' + parsedType.size + ' vs ' + num.bitLength())`

And maybe some others. It seems that those errors are pretty self descriptive. But do you suggest something like catching the errors and throwing a build error with that message?

FabianSimtlix: I think we should catch wrap them into a build error with a custom message. We probably don't want to expose that level of detail (uint, validations,etc) to the user

twtaylorbitgo: yeah, I think that's the ideal here. thanks guys.

src/coin/cgld/transactionBuilder.ts
twtaylorbitgo: Oh yeah I was not talking about behavior as much as style

```
switch (decodedType) {
      case TransactionType.StakingLock:
        this._stakingBuilder = new StakingBuilder()
          .type(StakingOperationsTypes.LOCK)
          .amount(transactionJson.value)
          .coin(this._coinConfig.name);
        break;
      case TransactionType.StakingVote:
        this._stakingBuilder = new StakingBuilder(transactionJson.data);
        break;
      default:
        super.setTransactionTypeFields(decodedType, transactionJson);
        break;
    }
```

vs

```
        switch (this._type) {
      case TransactionType.StakingLock:
        return this.buildLockStakeTransaction();
      case TransactionType.StakingVote:
        return this.buildVoteStakingTransaction();
    }
    return super.getTransactionData();
```

differ in how they return a default output.

twtaylorbitgo: this is not a blocker and feel free to merge the PR

PR #128, opened 2020-06-11
include signature in id/hash

FabianSimtlix:


src/coin/cgld/types.ts
```diff
@@ -188,7 +188,7 @@ export class CgldTransactionData implements EthLikeTransactionData {
       gasLimit: new BigNumber(bufferToHex(this.tx.gasLimit), 16).toString(10),
       value: this.tx.value.length === 0 ? '0' : new BigNumber(bufferToHex(this.tx.value), 16).toString(10),
       data: bufferToHex(this.tx.data),
-      id: addHexPrefix(bufferToHex(this.tx.hash())),
+      id: addHexPrefix(bufferToHex(this.tx.hash(true))),
```
marktoda, line 191: What if there is no signature yet? Shouldn't it check if there is `v`, `r`, and `s` set and set `useSignature` based on that?

marktoda: @FabianSimtlix ??

FabianSimtlix: I didn't saw this comment.
If the transaction is not signed yet `v` contains the chain id. `r`, `s` has default empty values.

marktoda: Ah so whats even the point of the `includeSignature` flag then

FabianSimtlix: Is for when you verify the signature, there you need a hash that does not contains the signature part

PR #132, opened 2020-06-15
Parse inputs, outputs and signatures for transactions

marktoda:
This commit adds the inputs, outputs, and signatures fields for send
transactions. It parses the transaction data to get these fields, and
exposes them through the same interface that other coins use. This will
help dependent libraries to properly decode transactions and understand
what is happening

Ticket: BG-22118

### DIFF
--- a/src/coin/eth/transaction.ts
+++ b/src/coin/eth/transaction.ts
@@ -8,7 +8,9 @@ import { InvalidTransactionError, SigningError } from '../baseCoin/errors';
 import { KeyPair } from './keyPair';
 import { EthLikeTransactionData, TxData } from './iface';
 import { EthTransactionData } from './types';
-import { classifyTransaction, hasSignature, toStringSig } from './utils';
+import { classifyTransaction, decodeTransferData, getToken, hasSignature, toStringSig } from './utils';
+
+const UNSUPPORTED_COIN_NAME = 'unsupported';
 
 export class Transaction extends BaseTransaction {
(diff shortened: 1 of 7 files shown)


src/coin/eth/transaction.ts
```diff
@@ -105,7 +133,9 @@ export class Transaction extends BaseTransaction {
     if (!keyPair.getKeys().prv) {
       throw new SigningError('Missing private key');
     }
-    this._transactionData.sign(keyPair);
+    await this._transactionData.sign(keyPair);
+    const txData = this._transactionData.toJson();
```
argjv, line 137: Is this for every type of transaction?

src/coin/eth/transaction.ts
```diff
@@ -105,7 +133,9 @@ export class Transaction extends BaseTransaction {
     if (!keyPair.getKeys().prv) {
       throw new SigningError('Missing private key');
     }
-    this._transactionData.sign(keyPair);
+    await this._transactionData.sign(keyPair);
+    const txData = this._transactionData.toJson();
+    this._signatures.push(toStringSig({ v: txData.v!, r: txData.r!, s: txData.s! }));
```
argjv, line 138: Is this done for every type of transaction? Or should we guard it with an `if`?

marktoda: Yes for every type of transaction -- this is the "external" signature which is required by the network for a tx to be valid

PR #134, opened 2020-06-16
Add token contract address to sends for CGLD

marktoda:
The transferBuilder accepts a coin name and susses out the token
contract idea from that using bitgo/statics if the coin is a token.
However, CGLD is *not* just a token in statics, but also a native
account coin. So this commit changes the builder function to be more
generic and accept token contract address for any type of coin that has
it.

Ticket: BG-22200


marktoda:
Blocked on an updated version of bitgojs from this PR: https://github.com/BitGo/BitGoJS/pull/768

marktoda:
Corresponding SDK change: https://github.com/BitGo/BitGoJS/pull/768

test/unit/coin/eth/transferBuilder.ts
```diff
@@ -85,12 +85,6 @@ describe('Eth send multi sig builder', function() {
       }).throw();
     });
 
-    it('should fail if a coin is a token but not an ERC20 token', () => {
-      should.throws(() => {
-        new TransferBuilder().coin('xlm:ETH-GBVOL67TMUQBGL4TZYNMY3ZQ5WGQYFPFD5VJRWXR72VA33VFNL225PL5');
-      }, 'There was an error using that coin as a transfer currency');
-    });
-
```
twtaylorbitgo, line 93: why is this test being removed

test/unit/coin/eth/transferBuilder.ts
```diff
@@ -133,12 +127,6 @@ describe('Eth send multi sig builder', function() {
       }).throw('Invalid expiration time');
     });
 
-    it('should fail if a coin is not an ERC20 instance', () => {
-      should.throws(() => {
-        new TransferBuilder().coin('eth');
-      }, 'There was an error using that coin as a transfer currency');
-    });
-
```
twtaylorbitgo, line 141: same

test/unit/coin/eth/transferBuilder.ts
marktoda: Because it's no longer true and should not be true! That's the point of this PR. I want to be able to set the coin for both tokens and not tokens and have the builder know whether or not to include the token contract address

```
builder.coin('eth')  // should build with a native send
builder.coin('wbtc') // should build with a token send
```

test/unit/coin/eth/transferBuilder.ts
marktoda: same as above

marktoda: Because it's no longer true and should not be true! That's the point of this PR. I want to be able to set the coin for both tokens and not tokens and have the builder know whether or not to include the token contract address

```
builder.coin('eth')  // should build with a native send
builder.coin('wbtc') // should build with a token send
```

test/unit/coin/eth/transferBuilder.ts
twtaylorbitgo: ah gotcha

test/unit/coin/eth/transferBuilder.ts
twtaylorbitgo: ah ok. I get it. could we add a test then for coverage of Line 45 in transferBuilder then?

```
    if (this._coin instanceof ContractAddressDefinedToken) {
```

marktoda: yep sounds good

PR #140, opened 2020-06-18
Add test for incorrect transaction id calculation

marktoda:
This commit adds a test to show the incorrect calculation of transaction
id for celo

Ticket: BG-22254


marktoda:
This was an actual transaction on chain, so something is different between the real txid calculatino and whatever we are doing here: https://alfajores-blockscout.celo-testnet.org/tx/0x9575af450d92763ea7c44e4a3930b2de4ec27bea3d81487dc6820ee4be042f58/internal_transactions

cc @FabianSimtlix as you know the most here

marktoda:
I thought we already fixed the `this.hash(true)` thing haha - what happened to that fix?

src/coin/eth/types.ts
```diff
@@ -54,7 +54,7 @@ export class EthTransactionData implements EthLikeTransactionData {
       gasLimit: new BigNumber(bufferToHex(this.tx.gasLimit), 16).toString(10),
       value: this.tx.value.length === 0 ? '0' : new BigNumber(bufferToHex(this.tx.value), 16).toString(10),
       data: bufferToHex(this.tx.data),
-      id: addHexPrefix(bufferToHex(this.tx.hash())),
+      id: addHexPrefix(bufferToHex(this.tx.hash(true))),
```
twtaylorbitgo, line 57: if we're gonna re-use this, it should be in a function with a unit test

src/coin/cgld/types.ts
```diff
@@ -188,7 +188,7 @@ export class CgldTransactionData implements EthLikeTransactionData {
       gasLimit: new BigNumber(bufferToHex(this.tx.gasLimit), 16).toString(10),
       value: this.tx.value.length === 0 ? '0' : new BigNumber(bufferToHex(this.tx.value), 16).toString(10),
       data: bufferToHex(this.tx.data),
-      id: addHexPrefix(bufferToHex(this.tx.hash())),
+      id: addHexPrefix(bufferToHex(this.tx.hash(true))),
```
twtaylorbitgo, line 191: comment below

PR #141, opened 2020-06-19
BGA-165 Remove source

ipelaez-simtlix:
We removed the source because we can get it from the signature


test/unit/coin/cgld/transactionBuilder/addressInitialization.ts
```diff
@@ -25,6 +24,7 @@ describe('An address initialization', () => {
       txJson.gasPrice.should.equal('1000000000');
       should.equal(txJson.nonce, 2);
       should.equal(txJson.chainId, 44786);
+      should.equal(txJson.from, testData.KEYPAIR_PRV.getAddress());
```
gonzaalvarez70, line 27: Maybe we could add a test where txJson.from should be undefined? I'm not sure if that's possible but it seems like a good test case.

PR #148, opened 2020-06-25
Use coin-specific wallet contracts

marktoda:
This commit updates the ETH-clones to use coin-specific wallet contracts
which verify signatures including an accurate network-specific modifier.
This ensures that signatures cannot be replayed across chains, even if
all other details are the same.

Ticket: BG-22425


marktoda:
How I got the bytecode:
```
Checkout https://github.com/bitgo/eth-multisig-v2 master
npm install -g solc@0.4.18
solcjs --bin ./contracts/coins/<coin>WalletSimple.sol ./contracts/Forwarder.sol ./contracts/ERC20Interface.sol
cat <coin>WalletSimple.bin
```

marktoda:
I added comments right where the bytecode is to be more contextual and to avoid cluttering the README with obscure info.

src/coin/cgld/utils.ts
```diff
@@ -66,5 +66,12 @@ export function getCommon(network: NetworkType): EthereumCommon {
   return common;
 }
 
+/*
```
marktoda, line 69: marker

marktoda:
like https://github.com/BitGo/bitgo-account-lib/pull/148/commits/30509c820877bf5504382ce49a2622b97d414d81#r445892200

PR #152, opened 2020-07-06
Splitting keypair class into secp and ed based keycurves

twtaylorbitgo:
From stellar seed still TBD

Ticket: BG-22623

### DIFF
--- a/src/coin/baseCoin/baseKeyPair.ts
+++ b/src/coin/baseCoin/baseKeyPair.ts
@@ -1,23 +1,31 @@
-import {AddressFormat} from "./enum";
-import {KeyPairOptions} from "./iface";
+import { AddressFormat } from './enum';
+import { KeyPairOptions } from './iface';
 
 /**
  * Base keys and address management.
  */
-export abstract class BaseKeyPair {
+export interface BaseKeyPair {
+  /**
+   * Build a set of keys from a prv
+   *
+   * @param {string} prv A raw private key
+   */
+  recordKeysFromPrivateKey(prv: string): void;
 
-    protected constructor(protected source?: KeyPairOptions) {}
+  /**
+   * Build a set of keys from a pub
+   *
+   * @param {string} prv A raw pub key
+   */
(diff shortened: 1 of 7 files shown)


src/coin/baseCoin/baseKeyPair.ts
```diff
@@ -0,0 +1,23 @@
+import {AddressFormat} from "./enum";
```
gonzaalvarez70, line 1: I think that you don't have eslint enabled, it was configured within the pre-commit step so I'm not sure what happened. But I  just noticed that there are a few indentations/spacing issues.

src/coin/baseCoin/baseKeyPair.ts
```diff
@@ -0,0 +1,23 @@
...
+import {KeyPairOptions} from "./iface";
+
+/**
+ * Base keys and address management.
+ */
+export abstract class BaseKeyPair {
```
argjv, line 7: This should be an interface

src/coin/baseCoin/ed25519KeyPair.ts
```diff
@@ -0,0 +1,5 @@
+import {BaseKeyPair} from "./baseKeyPair";
+
+export abstract class Ed25519KeyPair extends BaseKeyPair {
```
argjv, line 3: Class JS doc

src/coin/baseCoin/baseKeyPair.ts
twtaylorbitgo: oh that's funny. thanks

src/coin/baseCoin/baseKeyPair.ts
twtaylorbitgo: I'll make it an interface for now

In terms of composition we need a base class for the KeyPair functionality though and should tack on any aspects of Keys via interfaces if this gets out of hand

Things like ```IFromXlmSeed```, etc.

PR #153, opened 2020-07-07
Change CGLD to CELO

marktoda:
This commit changes the name of the CGLD coin to CELO, to match the name
change that the network performed a couple weeks ago. One more change
will need to be made once @bitgo/statics changes the name as well.

This commit also updates the version to a new major - 2.0.0 as the name
change for celo is a breaking change

Ticket: BG-22663


marktoda:
blocked on https://github.com/BitGo/BitGoJS/pull/785

PR #158, opened 2020-07-09
Deprecate repo

tylerlevine:
Remove source code which has been moved to SDK monorepo and update README with link to new repo.

Ticket: BG-22556


tylerlevine:
Yes it should be archived, but I don't have permission to do that myself.

If this needs to be reversed, run these commands:
```
git revert a7f30609f5d8af78e8b5dcc756af818430c20e9e
git revert b1d226668503e9f5b4cd79dbc3e4f88ff8d5856e
```

and everything should be back to where we started. Hopefully this is not necessary.

tylerlevine:
I've also just pushed a tag called `last-commit` which points to the commit before everything was deleted. If we really need to, we can branch off that tag to make additional changes.

